PR #8, opened 2018-04-24
[INTERNAL] generateManifestBundle: Fix unit test issue in CI environments

RandomByte:
In some CI environments (Travis, Alpine Linux in Docker) we saw the "Build application.b with manifestBundler" test failing with the following error:

> 'Invalid filename: manifest.json'
> Error code: ELIFECYCLE

This PR fixes that error in those environments


RandomByte:
Well, the error mentioned in my initial comment actually disappeared. I guess it was related to a dependency causing and fixing this. But that's only a guess.

Closing this PR for now.

PR #23, opened 2018-06-18
[FEATURE] Add module type

RandomByte:
Prerequisite for third-party shimming based on https://github.com/SAP/ui5-tooling/pull/5

Depends on https://github.com/SAP/ui5-project/pull/46

Example implementation: https://github.com/RandomByte/openui5-parallax/pull/1/files#diff-f7cafeb3d075b6c5d28c1102cec23f52R14

### DIFF
--- a/lib/types/module/ModuleBuilder.js
+++ b/lib/types/module/ModuleBuilder.js
@@ -1,0 +1,12 @@
+const AbstractBuilder = require("../AbstractBuilder");
+
+class ModuleBuilder extends AbstractBuilder {
+	constructor({resourceCollections, project, parentLogger}) {
+		super({project, parentLogger});
+
+		// All available library tasks in execution order
+		this.availableTasks = [];
+	}
+}
+
+module.exports = ModuleBuilder;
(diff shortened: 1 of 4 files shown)


lib/types/module/ModuleBuilder.js
```diff
@@ -0,0 +1,27 @@
...
+class ModuleBuilder extends AbstractBuilder {
+	constructor({resourceCollections, project, parentLogger}) {
+		super({project, parentLogger});
+
+		// All available library tasks in execution order
+		this.availableTasks = [
+			"uglify"
+		];
+
+		this.addTask("uglify", () => {
```
matz3: Do we really want to always uglify?
I would expect that modules from the npm registry are already optimized.

RandomByte: Done

RandomByte:
I think we should tackle increasing the test coverage of the types separately. While the "module" type has a coverage of ~20% now, the others have similarly bad ~50%. There are no dedicated tests for the types yet.

RandomByte:
@matz3 ping

PR #26, opened 2018-06-21
Create and integrate a new task 'generateLibraryManifest'

codeworrior:
The task creates a manifest.json file for libraries that don't have one.
It collects the necessary information from the project structure
(supported themes, versions of dependencies), from the .library file
(name, dependencies, supported themes, thirdparty components, i18n
information...) and from an initLibrary call inside the library.js file
(implemented controls, elements, types and interfaces, noLibraryCSS
flag).

An existing manifest.json won't be modified or overwritten.


codeworrior:
Note: the test failures are due to a bug in the ui5-fs project. The bug has been fixed with https://github.com/SAP/ui5-fs/pull/11, which might be part of 0.0.2.

RandomByte:
There's yet no ui5-fs release containing that fix. Will create one later today

matz3:
@codeworrior LGTM, I just fixed some minor copy & paste leftovers

lib/tasks/bundlers/generateLibraryManifest.js
```diff
@@ -0,0 +1,60 @@
...
+	const combo = new ReaderCollectionPrioritized({
+		name: `libraryManifestGenerator - prioritize workspace over dependencies: ${options.projectName}`,
+		readers: [workspace, dependencies]
+	});
+	return combo.byGlob("/**/*.{js,json,xml,html,properties,library,less,css,theming,theme}").then((resources) => {
+		// Find all libraries and create a manifest.json file
+		return workspace.byGlob("/resources/**/.library").then((libraryIndicatorResources) => {
+			if (libraryIndicatorResources.length < 1) {
+				// No library found - nothing to do
+				log.error(`Could not find a ".library" file for project ${options.projectName}. Skipping library manifest generation.`);
```
matz3: Do we want this to be an error?
I think most of the libraries developed in the wild doesn't have a `.library` file as it wasn't used at all except by the maven tooling.

codeworrior: If we take "manifest first for libraries" serious, we need at least to enforce a `manifest.json`. So we could check for that and if there's a `manifest.json` then the `.library` becomes optional.

But it doesn't have to be this task that complains about it.

RandomByte: More likely the job of the [LibraryFormatter](https://github.com/SAP/ui5-builder/blob/master/lib/types/library/LibraryFormatter.js#L61).

The ApplicationFormatter already [reads an applications manifest](https://github.com/SAP/ui5-builder/blob/master/lib/types/application/ApplicationFormatter.js#L16), though it doesn't insist on it.

codeworrior:
any chance to get another review?

codeworrior:
merci :)

codeworrior:
@Matz3: when doing this PR, I didn't know about the best practice of exporting everything in `index.js`. Shouldn't the `generateLibraryManifest` task also be added there?

RandomByte:
For consistency reasons I would say yes.

PR #31, opened 2018-06-22
Fix #30 by using a for loop instead of .forEach

### LINKED ISSUE #30: Builder (kind of) fails when routes are defined as an object
## Expected Behavior
Defining routes in the manifest.json as an object instead of an array should be perfectly "legal" (see [sap.ui.routing.Router](https://github.com/SAP/openui5/blob/master/src/sap.ui.core/src/sap/ui/core/routing/Router.js#L36)). 

## Current Behavior
The build throws errors, but "succeeded".
The only difference between a "bad build" and a "good build" (when using an array) seems to be the the manifest itself (d'oh!). 

## Steps to reproduce the issue
1. Create a project
2. add Routes to you manifest
```json
      "routes": {
        "start": {
          "target": "return_list",
          "pattern": ""
        },
        "create_request": {
          "target": "create_request",
          "pattern": "create/{indices}"
        }
      },
```
3. `ui5 build`
4. watch it crash

## Context
- OS/Platform: fedora 27 linux 
- Node.js Version: 8.11.2
- npm Version: 6.10
- Other information: ui5/cli v0.0.2

## Affected components
- [X] [ui5-builder](https://github.com/SAP/ui5-builder)
- [?] [ui5-server](https://github.com/SAP/ui5-server)
- [?] [ui5-cli](https://github.com/SAP/ui5-cli)
- [?] [ui5-fs](https://github.com/SAP/ui5-fs)
- [?] [ui5-project](https://github.com/SAP/ui5-project)
- [?] [ui5-logger](https://github.com/SAP/ui5-logger)

## Stack trace/log output
```
ERR! ComponentAnalyzer TypeError: routing.routes.forEach is not a function
ERR! ComponentAnalyzer     at ComponentAnalyzer._analyzeManifest (/home/masch/code/customer/Project/node_modules/@ui5/builder/lib/lbt/analyzer/ComponentAnalyzer.js:117:19)
ERR! ComponentAnalyzer     at ComponentAnalyzer.analyze (/home/masch/code/customer/Project/node_modules/@ui5/builder/lib/lbt/analyzer/ComponentAnalyzer.js:56:9)
ERR! ComponentAnalyzer     at <anonymous>
ERR! ComponentAnalyzer     at process._tickCallback (internal/process/next_tick.js:188:7)
```
https://github.com/SAP/ui5-builder/blob/master/lib/lbt/analyzer/ComponentAnalyzer.js#L117

themasch:
> iterate over routes using a for loop instead of .forEach so we can iterate over objects as well

I figured I might as well create a PR for this issue. afaik this should be the "cleanest" solution for this problem the inner body does not care about the routes name anyway. 

It hurts me a little bit to remove functional syntax and the fat arrow, but I guess its for the greater good ;)

An alternative would be to use `Object.values(routing.routes).forEach` but that would give no access to the route name at all. With the for loop theres still the `key` if one needs the name. 

Additionally I guess this is faster than Object.values.

### DIFF
--- a/lib/lbt/analyzer/ComponentAnalyzer.js
+++ b/lib/lbt/analyzer/ComponentAnalyzer.js
@@ -113,21 +113,40 @@ class ComponentAnalyzer {
 
 		let routing = ui5.routing;
 		if ( routing ) {
-			// console.log("routing: ", routing);
-			routing.routes.forEach( (route) => {
-				let target = routing.targets[route.target];
-				if ( target && target.viewName ) {
-					let module = ModuleName.fromUI5LegacyName(
-						(routing.config.viewPath ? routing.config.viewPath + "." : "") +
-						target.viewName, ".view." + routing.config.viewType.toLowerCase() );
-					log.verbose("converting route to view dependency ", module);
-					// TODO make this a conditional dependency, depending on the pattern?
-					info.addDependency(module);
+			if (Array.isArray(routing.routes)) {
+				routing.routes.forEach((route) => this._visitRoute(route, routing, info));
+			} else {
+				for (let key in routing.routes) {
+					if (!routing.routes.hasOwnProperty(key)) {
+						continue;
+					}
+					const route = routing.routes[key];
+					this._visitRoute(route, routing, info);
 				}
-			});
+			}
 		}
 
 		return info;
+	}
+
+	/**
+	 * called for any route, this adds the view used by a route as a dependency.
+	 *
+	 * @param {object} route the single route
+	 * @param {object} routing the full routing object from the manifest
+	 * @param {ModuleInfo} info  ModuleInfo object that should be enriched
+	 * @private
+	 */
+	_visitRoute( route, routing, info ) {
+		const viewPath = routing.config.viewPath ? routing.config.viewPath + "." : "";
+		const viewType = routing.config.viewType.toLowerCase();
+		const target = routing.targets[route.target];
+		if ( target && target.viewName ) {
+			const module = ModuleName.fromUI5LegacyName(viewPath + target.viewName, ".view." + viewType);
+			log.verbose("converting route to view dependency ", module);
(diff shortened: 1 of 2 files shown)


codeworrior:
Yes, order of routes is currently not of any interest for the dependency analysis.

matz3:
I would prefer a more explicit handling of both types as it might be hard to understand for others. At least I was at first thinking that only objects are supported now. But if I'm the only one, I don't mind 😄

themasch:
@codeworrior oh sure. I'll edit the commit message and force push. 

@matz3 I could do something like 
```js
if (Array.isArray(routing.routes)) {
  routing.routes.forEach(this.visitRoute.bind(this, info))
} else {
  for ( let key in routing.routes) {
    this.visitRoute(info, routing.routes[key])
  }
}
``` 
and move the loop body to a method (`visitRoute` in this example). 

That way the handling of arrays would not change fromt he current version and objects would be handled like in this pr. 

Another way would be to use `Object.values(routing.routes).forEach(…)`.

matz3:
Looks good to me 👍 
@codeworrior @RandomByte what's your opinion?

themasch:
Oh sure, :+1: for the arrow function :D 

Should I restructure the PR to implement it that way?

RandomByte:
> Should I restructure the PR to implement it that way?

Yes please. I think @codeworrior will also be fine with that 🙂

codeworrior:
I had fixed those two issues in a branch of our internal predecessor repo already, but that branch didn't make it into the public repos.

That's how I fixed the second:

```javascript
	// merge target config with default config
	let config = Object.assign({}, routing.config, target);
	let module = ModuleName.fromUI5LegacyName(
		(config.viewPath ? config.viewPath + "." : "") +
		config.viewName, ".view." + config.viewType.toLowerCase() );
```

If you don't mind, I can push the two fixes on top of your change. It might take some time until I can migrate the old branch to the `ui5-builder` repo.

PR #42, opened 2018-06-28
[FEATURE] Add JSDoc build functionalities

codeworrior:
Note: the code is full of console.log just to track down an unexpected  exit of the Node.js process.


RandomByte:
By writing a resource, the FileSystem adapter drains its content stream. Therefore the `.clone` workaround makes sense. However, I'll look into a way to have the FS adapter to provide a new stream.

RandomByte:
I think 1. is closer to the expected behavior, I'll try to prepare a ui5-fs PR for that soon.

RandomByte:
Started working on the discussed ui5-fs fix in https://github.com/SAP/ui5-fs/pull/22

RandomByte:
Fix for blocker got merged: SAP/ui5-fs#22

RandomByte:
Rebased

RandomByte:
### Implementation Concept
[JSDoc generation UI5 Tooling integration.pdf](https://github.com/SAP/ui5-builder/files/2905832/JSDoc.generation.UI5.Tooling.integration.pdf)

### Additional Implementation Details

- Preferably add a "ui5 build jsdoc” command defining all JSDoc related tasks as enabled by default
    - It is still possible to activate single JSDoc related tasks in any other build using the corresponding CLI parameter
- Enhance [Custom Task|https://github.com/SAP/ui5-project/blob/master/docs/BuildExtensibility.md#custom-task-extension] implementation for defining "final" tasks (lower part of the chart) and positioning as “first/last” task as there might not be any active "final" task that could be used for relative positioning
- Add ui5.yaml configuration for JSDoc excludes (like thirdparty- or very big, not relevant resources). Grunt JSDoc build currently does not have any exclude handling. Maven JSDoc build reads this information from .library
- Processor returns array of resources but also writes to fs? (yes, at least for jsdoc config, but maybe different location?) (maybe different dest location?)

CC: @tommyvinhlam

test/lib/tasks/generateJsdoc.js
```diff
@@ -1,9 +1,84 @@
...
+	t.context.tmpStub.callsArgWithAsync(1, {message: "Dir creation failed"}, "some/path");
+
+	const res = await t.throws(generateJsdoc._createTmpDir("some.namespace"));
+
+	t.deepEqual(t.context.tmpStub.callCount, 1, "Tmp dir is called once");
+	t.deepEqual(t.context.tmpStub.getCall(0).args[0].prefix, "ui5-tooling-tmp-jsdoc-some.namespace-");
+	t.deepEqual(res, {message: "Dir creation failed"}, "Dir creation failed");
+});
+
+test.serial("createTempDirs", async (t) => {
```
RandomByte: create**Tmp**Dirs

RandomByte:
Not sure anymore whether the project name should be included in the temporary path directly. At least we should remove all alpha-num characters to prevent directory traversal

RandomByte:
b93d4bd depends on https://github.com/SAP/ui5-fs/pull/112

RandomByte:
**TODO**: Document new JSDoc exclude configuration in UI5 Project

**Example from sap.ui.core:**
```yaml
---
specVersion: "0.1"
type: library
metadata:
  name: sap.ui.core
  copyright: |-
   OpenUI5
    * (c) Copyright 2009-${currentYear} SAP SE or an SAP affiliate company.
    * Licensed under the Apache License, Version 2.0 - see LICENSE.txt.
builder:
  jsdoc:
    excludes:
    - "jquery-*"
    - "sap-ui-*"
    - "sap/ui/debug/**"
    - "sap/ui/qunit/**"
    - "sap/ui/thirdparty/**"
```

RandomByte:
Still not sure whether JSDoc exclude patterns should include the virtual base path (`/resources` / `/test-resources`) to allow differentiation between the two (or more). Currently they are designed to only work within `/resources` (just like the JSDoc build itself).

This came to my mind when thinking about the definition of project-wide exclude patterns as requested in https://github.com/SAP/ui5-tooling/issues/124. They should probably include the virtual paths to allow excluding resources of `/test-resources` for example.

So even though the JSDoc build itself is only working on `/resources`, having the same flavor of include/exclude patterns (that include the virtual base path) in the `ui5.yaml` might make sense.

matz3:
@RandomByte I would align this with the bundler, which also doesn't contain `resources` or `test-resources` in the pattern/filters but implicitly uses `resources`.

RandomByte:
We had another discussion and tend to keep JSDoc exclude patterns aligned with the bundler patterns while use absolute virtual paths (`/resources/pony/**`) for more general excludes.

RandomByte:
Depends on https://github.com/SAP/ui5-fs/pull/113

RandomByte:
Rebased

codeworrior:
I've tried it out on OpenUI5 (together with the pending change there) and it worked quite well!

Issue #36, opened 2018-06-25
Configuration in ui5.yaml file for the createDebugFiles Task


stopcoder:
## Expected Behavior
I would like to be able to write additional configuration in the ui5.yaml file for the createDebugFiles task to:
 - skip a source file from being created with debug file
   - Example: there's a source file in our library, let's call it bootstrap.js. The file is used only for development time. Once the library is built, the content of this file is replaced completely but saved still under the same path. We also create a bundle for being used under debug mode. This means that we create two bundles, bootstrap.js and bootstrap-dbg.js. We need a configuration to skip the creation of the bootstrap-dbg.js otherwise the dbg file is created by copy the content of the bootstrap.js bundle which is different than the bootstrap-dbg.js bundle.
 - specify the bundles which are created in createBundle task to be created with debug files
   - some of our bundle files should also be created with debug files. Because they don't appear as source files but only be created in the createBundle task, they are currently not being considered in the createdDebugFiles task.

## Current Behavior
Debug files are created for each and every javascript source file. There's no way to add additional and to skip the existing source files.

## Affected components
- [x] [ui5-builder](https://github.com/SAP/ui5-builder)
- [ ] [ui5-server](https://github.com/SAP/ui5-server)
- [ ] [ui5-cli](https://github.com/SAP/ui5-cli)
- [ ] [ui5-fs](https://github.com/SAP/ui5-fs)
- [ ] [ui5-project](https://github.com/SAP/ui5-project)
- [ ] [ui5-logger](https://github.com/SAP/ui5-logger)

RandomByte:
After discussion with @stopcoder:

**Re 1:** A possible solution would be to add a configuration option for the pattern that is handed over to `createDebugFiles` in Application- and LibraryBuilder

**Re 2:** A possible solution would be to add a bundle configuration parameter like "createDebugFile" to automatically create a `-dbg.js` version of a bundle.  
This seems to be required for all bundles that do not have any fallback logic in the UI5 framework when running an application in sap-ui-debug mode (to be confirmed @codeworrior).

Moreover, the `-dbg.js` version should not be minified. But when splitting a bundle into multiple parts (via `numberOfParts` config), that splitting should happen evenly for the minified sourced (independently from the `-dbg.js` sources).


Also @codeworrior: Can you shed some light into how something like this worked with the maven tooling?

RandomByte:
@matz3 mentioned that there is actually a fallback logic from `x-dbg.js` to `x.js`.

[...]

PR #43, opened 2018-06-28
[INTERNAL] Task createDebugFiles checks path before writes

stopcoder:
- The `debugFileCreator` processor checks whether the debug file path
  is used and only writes the content to the path when it's not
  used yet.

- The debug file path can be used, for example, by the `createBundle`
  task which may create a specific bundle for the debug version of
  another bundle. In this case, the `debugFileCreator` shouldn't
  overwrite the created debug bundle.

### DIFF
--- a/lib/processors/debugFileCreator.js
+++ b/lib/processors/debugFileCreator.js
@@ -1,4 +1,5 @@
 const copier = require("./resourceCopier");
+const util = require("util");
 
 /**
  * Creates *-dbg.js files for all supplied resources.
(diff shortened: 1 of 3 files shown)


stopcoder:
This solves the first point in issue #36

stopcoder:
Build time comparison with [openui5-sample-app](https://github.com/SAP/openui5-sample-app/commit/c87c36623fdc8152e06c82edbffc8eb0b02196a0):

**master**
```
time ui5 build -all
[...]
ui5 build --all 71.30s user 6.79s system 130% cpu 59.818 total
```

**debug_file_check_existence_before_write**
```
time ui5 build -all
[...]
ui5 build --all 71.02s user 6.80s system 130% cpu 59.563 total
```

### **Summary**
The extra check of path existence in the workspace **doesn't prolong** the build time.

RandomByte:
Could you squash those commits into one?

stopcoder:
Somehow the squash doesn't work well, I am working on it.

stopcoder:
I forgot `git add -u` after a soft reset 😅

Now everything should be OK.

lib/processors/debugFileCreator.js
```diff
@@ -8,12 +9,41 @@ const copier = require("./resourceCopier");
...
+
+	const stat = util.promisify(fs.stat);
+
+	return Promise.all(
+		resources.map((resource) => {
+			// check whether the debug resource path is already used in the
+			// previous tasks
+			return stat(resource.getPath().replace(options.pattern, options.replacement))
+				.then(
+					// if the file can be found, it should be filtered out from createing debug file
```
RandomByte: createing => creating

lib/processors/debugFileCreator.js
```diff
@@ -8,12 +9,41 @@ const copier = require("./resourceCopier");
...
+						if (err.code === "ENOENT") {
+							// if the file can't be found, it should be included in creating debug file
+							return resource;
+						}
+						// if it's other error, forward it
+						throw err;
+					}
+				);
+		})
+	).then((stats) => {
```
RandomByte: What about `results` instead of `stats`, as the array doesn't contain any stat info?

lib/processors/debugFileCreator.js
```diff
@@ -8,12 +9,41 @@ const copier = require("./resourceCopier");
...
+		// filter out the resouces whose debug source path is already used
+		return stats.filter((stat) => {
+			return !!stat;
+		});
+	}).then((filteredResource) => {
```
RandomByte: `filteredResource` => `filteredResources`

lib/processors/debugFileCreator.js
stopcoder: done

lib/processors/debugFileCreator.js
stopcoder: done

lib/processors/debugFileCreator.js
stopcoder: done

PR #49, opened 2018-07-05
[INTERNAL] Remove greenkeeper

RandomByte:
And update `package-lock.json` which got weird with 000a6fee49555cb266a1c575cde719e1091d1066 (EDIT: already done with #48, got removed from this commit)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,5 @@
 language: node_js # don't install any environment
 
 node_js:
-- '8'
-- '10'
-
-before_install: npm install -g greenkeeper-lockfile
-install: if [[ "$TRAVIS_BRANCH" =~ ^greenkeeper.*$ ]] || [[ `npm -v` == 5* ]]; then
-  echo "npm install" && npm install; else echo "npm ci" && npm ci; fi
-
-before_script: greenkeeper-lockfile-update
-after_script:
-- npm run report-coveralls
-- greenkeeper-lockfile-upload
-
-env:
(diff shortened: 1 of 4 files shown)


codeworrior:
> And update package-lock.json which got weird with 000a6fe

That's a question that I anyhow had: isn't there a way to check the package-lock.json for plausibility, either locally or as part of the Travis builds? I always felt uncomfortable when adding that file to a change as the diff usually was too big to understand it. 

Maybe something like
```
  cp package-lock.json package-lock-candidate.json
  git checkout -f package-lock.json 
  npm install
  diff -U 3 package-lock-candidate.json package-lock.json
```
  
But maybe I just have to ban `npm update` from my mental toolkit and everything will be fine.

RandomByte:
@codeworrior This would make sense as part of the Travis build. I think the odd change in 000a6fee49555cb266a1c575cde719e1091d1066 might have been caused by the usage of an older npm version.

But what's the purpose of `git checkout -f package-lock.json`?

codeworrior:
Well, that would work only locally before commit. Idea was to revert any local changes, let NPM recalculate the package-lock.json and check if there's anything unexpected in the diff.

RandomByte:
I created https://github.com/SAP/ui5-tooling/issues/22 to track possible CI enhancements.

RandomByte:
Also see https://github.com/SAP/ui5-logger/pull/15#issuecomment-403040728

PR #65, opened 2018-07-17
[FIX] generateLibraryManifest: i18n/css handling

matz3:


test/lib/tasks/generateLibraryManifest.js
```diff
@@ -0,0 +1,484 @@
...
+				<version>2.0.0</version>
+
+				<documentation>Test Lib</documentation>
+
+			</library>
+		`,
+		project: t.context.workspace._project
+	}));
+
+	await assertCreatedManifest(t, {
```
codeworrior: much nicer testing :clap:

lib/processors/manifestCreator.js
```diff
@@ -399,9 +399,9 @@ async function createManifest(libraryResource, libBundle, descriptorVersion, _in
 				let i18nElement = findChild(libraryAppData, "i18n");
 				if ( i18nElement ) {
 					let i18n = i18nElement._;
-					if ( "false".equals(i18n) ) {
```
codeworrior: Peter suggested to polyfill `Object.prototype.equals` to help :older_man: with :coffee: background.

test/lib/tasks/generateLibraryManifest.js
```diff
@@ -0,0 +1,484 @@
...
+	return resourceFactory.createAdapter({
+		fsBasePath: path.join(__dirname, "..", "..", "fixtures", "sap.ui.core-evo", "main", "src"),
+		virBasePath: "/resources"
+	});
+}
+
+async function assertCreatedManifest(t, oExpectedManifest) {
+	const {workspace, dependencies, resources} = t.context;
+
+	await Promise.all(resources.map((resource) => workspace.write(resource)));
```
codeworrior: Just out of curiosity: why is writing needed although the resources have been created already? To make globbing work?

lib/processors/manifestCreator.js
matz3: Hahaha 😄

test/lib/tasks/generateLibraryManifest.js
matz3: Yes. The resources are created without the context of the workspace. So by just creating them, they can't be found in the task.
When the processor would be tested standalone, then the plain resources could be handed.

When writing the test I noticed that this isn't really straight forward so I think we could improve the fs API to be more elegant.

test/lib/tasks/generateLibraryManifest.js
matz3: I only added the relevant tests for the changed code, but I want to add more cases to cover the complete task / processor in here.

The only thing I don't like is passing the `project`, as I'm not sure it will be the same when using e.g. the CLI.

PR #67, opened 2018-07-27
[FEATURE] BundleBuilder: support modules using ES6 with usePredefineCalls

petermuessig:
When using the bundle option usePredefineCalls modules using ES6
features such as arrow functions, classes, ... are not supported
and the bundler will break as the mozilla-ast of uglify-es is not
ES6 ready.

### DIFF
--- a/lib/lbt/bundle/Builder.js
+++ b/lib/lbt/bundle/Builder.js
@@ -5,6 +5,7 @@
 const uglify = require("uglify-es");
 const {pd} = require("pretty-data");
 const esprima = require("esprima");
+const escodegen = require("escodegen");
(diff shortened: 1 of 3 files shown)


petermuessig:
@codeworrior any objections on the pull request - the performance decrease IMO can be ignored for the moment until we have a better solution. Using it on a trivial library shows code generation times of about 0-3 ms for standard library resources!

codeworrior:
I did some measurements (`ui5 build --all`)

Project | before PR | with PR
--------- | ---------- | --------
openui5-sample-app | 0:52 | 0:52 
openui5/testsuite | 3:58 | 4:05

Neglectable, I would say.

PR #197, opened 2019-03-04
[INTERNAL] ComponentAnalyzer Test: Remove additional promise

matz3:
Promise.reject already creates a promise so the async function is not
needed.
This caused the error log to be displayed as last entry in the test
output.


matz3:
Already fixed via https://github.com/SAP/ui5-builder/pull/164

PR #216, opened 2019-03-24
[FEATURE] clear build folder

mauriciolauffer:
This PR address the following issue: #45

## Pull Request Checklist

### DIFF
--- a/lib/builder/builder.js
+++ b/lib/builder/builder.js
@@ -1,4 +1,5 @@
-const rimraf = require("rimraf");
+const {promisify} = require("util");
+const rimraf = promisify(require("rimraf"));
 const log = require("@ui5/logger").getGroupLogger("builder:builder");
 const resourceFactory = require("@ui5/fs").resourceFactory;
 const MemAdapter = require("@ui5/fs").adapters.Memory;
@@ -147,7 +148,7 @@ module.exports = {
 	 * @param {Object} parameters Parameters
 	 * @param {Object} parameters.tree Dependency tree
 	 * @param {string} parameters.destPath Target path
-	 * @param {boolean} [parameters.clearDestPath=false] Decides whether project should clear the target path before build
+	 * @param {boolean} [parameters.clearDest=false] Decides whether project should clear the target path before build
 	 * @param {boolean} [parameters.buildDependencies=false] Decides whether project dependencies are built as well
 	 * @param {boolean} [parameters.dev=false] Decides whether a development build should be activated (skips non-essential and time-intensive tasks)
 	 * @param {boolean} [parameters.selfContained=false] Flag to activate self contained build
@@ -158,7 +159,7 @@ module.exports = {
 	 * @returns {Promise} Promise resolving to <code>undefined</code> once build has finished
 	 */
 	build({
-		tree, destPath, clearDestPath = false,
+		tree, destPath, clearDest = false,
 		buildDependencies = false, dev = false, selfContained = false, jsdoc = false,
 		includedTasks = [], excludedTasks = [], devExcludeProject = []
 	}) {
@@ -248,10 +249,10 @@ module.exports = {
 					tasks: projectTasks,
 					project,
 					parentLogger: log
-				}).then(() => {
-					return new Promise((resolve) => {
-						return (clearDestPath) ? rimraf(destPath, resolve) : resolve();
-					});
+				}).then(async () => {
+					if (clearDest) {
+						await rimraf(destPath);
+					}
 				}).then(() => {
 					log.verbose("Finished building project %s. Writing out files...", project.metadata.name);
 					buildLogger.completeWork(1);
--- a/lib/builder/builder.js
+++ b/lib/builder/builder.js
@@ -1,3 +1,4 @@
+const rimraf = require("rimraf");
 const log = require("@ui5/logger").getGroupLogger("builder:builder");
 const resourceFactory = require("@ui5/fs").resourceFactory;
 const MemAdapter = require("@ui5/fs").adapters.Memory;
@@ -246,6 +247,10 @@ module.exports = {
 					tasks: projectTasks,
 					project,
 					parentLogger: log
+				}).then(() => {
+					return new Promise((resolve) => {
+						rimraf(destPath, resolve);
+					});
 				}).then(() => {
 					log.verbose("Finished building project %s. Writing out files...", project.metadata.name);
 					buildLogger.completeWork(1);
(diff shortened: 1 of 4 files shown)


RandomByte:
I would rather not always clean the dist directory. There can be cases where you just want to overwrite specific files with a smaller and faster build (i.e. updating some files without re-doing the whole build).

An example for this might be the new JSDoc build and how it is used in the OpenUI5 testsuite:
https://github.com/SAP/openui5/blob/master/src/testsuite/package.json#L18-L19

You have to execute the full build one time. After that only a subset, of the build tasks is required to update the JSDoc. This speeds up the build time by about one minute in this example.

So maybe we could add an option to activate cleaning the dist directory before build? In that case I would also rather locate this in the UI5 CLI than in the UI5 Builder project, since I would not expect a function that is just called `build` to also remove things from my disk (remember that the UI5 Builder is not only used by the CLI but directly in other node scripts!).

mauriciolauffer:
Hi @RandomByte.

I've defined it as a parameter, false by default. I've added a test case for it. Also, I've moved rimraf from devDependencies to dependencies.

mauriciolauffer:
I'm not sure whether it should be a command or an option in ui5-cli. I'm more inclined to use it as an option... What do you think is better?

mauriciolauffer:
PS: I tried to create it as a task, but I couldn't find a way to pass the destPath as parameter to the task without changing too many files...

lib/builder/builder.js
```diff
@@ -246,6 +248,10 @@ module.exports = {
 					tasks: projectTasks,
 					project,
 					parentLogger: log
+				}).then(() => {
```
RandomByte: For readability, and also to catch any errors that rimraf might raise, rather wrap rimraf using [promisify](https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original):
```js
const {promisify} = require("util");
const rimraf = promisify(require("rimraf"));
```

Then use an async function like this:
```js
				}).then(async () => {
					if (clearDestPath) {
						await rimraf(destPath);
                    }
				}).then(() => {
				...
```

package-lock.json
```diff
@@ -1,8870 +1,8870 @@
...
-        "escape-string-regexp": "^1.0.5",
-        "supports-color": "^5.3.0"
-      }
-    },
-    "chardet": {
-      "version": "0.7.0",
-      "resolved": "https://registry.npmjs.org/chardet/-/chardet-0.7.0.tgz",
-      "integrity": "sha512-mT8iDcrh03qDGRRmoA2hmBJnxpllMR+0/0qlzjqZES6NdiWDcZkCNAk4rPFZ9Q85r27unkiNNg8ZOiwZXBHwcA==",
-      "dev": true
-
```
RandomByte: The `package-lock.json` change is pretty huge due to whitespace changes. I suspect that this is fine in this case, since it now uses tabs like our `package.json`.

Can you please let me know what version of npm you are using (`npm --version`)? For `6.9.0` I know for sure that this change is expected and fine.

test/lib/builder/builder.js
```diff
@@ -105,6 +105,35 @@ test("Build application.a [dev mode]", (t) => {
 	});
 });
 
+test("Build application.a and clear target path [dev mode]", (t) => {
```
RandomByte: This test does execute the newly added lines, but doesn't really test whether something got removed.

I would rather mock `rimraf` for this test and check whether it got called with the correct argument.  But we can add that test in a follow up change from our side, so feel free to leave this one as is.

**Example for mock-testing:**
https://github.com/SAP/ui5-builder/blob/3bf0f56097163ad04b3dfd1e9efab40ccd3607b5/test/lib/types/library/LibraryFormatter.js#L168-L174

lib/builder/builder.js
```diff
@@ -146,6 +147,7 @@ module.exports = {
 	 * @param {Object} parameters Parameters
 	 * @param {Object} parameters.tree Dependency tree
 	 * @param {string} parameters.destPath Target path
+	 * @param {boolean} [parameters.clearDestPath=false] Decides whether project should clear the target path before build
```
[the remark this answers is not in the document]
mauriciolauffer: Renamed.

lib/builder/builder.js
mauriciolauffer: Wrapped.

package-lock.json
mauriciolauffer: Yep, npm v6.9.0

test/lib/builder/builder.js
mauriciolauffer: Yes, it does test whether the folder has been deleted. Build is called twice. The first one creates a sub folder called "/rubbish-should-be-deleted" and clearDest=false. The second, builds with clearDest=true.

If you call the second build with clearDest=false, test will fail due to **assert.directoryDeepEqual(destPath, expectedPath)**

test

lib/builder/builder.js
```diff
@@ -146,6 +148,7 @@ module.exports = {
 	 * @param {Object} parameters Parameters
 	 * @param {Object} parameters.tree Dependency tree
 	 * @param {string} parameters.destPath Target path
+	 * @param {boolean} [parameters.clearDest=false] Decides whether project should clear the target path before build
```
RandomByte: We discussed this internally and would favor renaming this feature to "**clean**Dest". 

Other tools seem to call this functionality "clean" rather than "clear" as well:  
- Maven: https://maven.apache.org/plugins/maven-clean-plugin/
- Grunt: https://github.com/gruntjs/grunt-contrib-clean
- make: https://www.gnu.org/software/make/manual/html_node/Cleanup.html

lib/builder/builder.js
```diff
@@ -246,6 +249,10 @@ module.exports = {
 					tasks: projectTasks,
 					project,
 					parentLogger: log
+				}).then(async () => {
+					if (clearDest) {
```
RandomByte: Also feedback from our internal discussion: It might make sense to move the dest removal part to *before* the build is started. I.e. Line [273](https://github.com/SAP/ui5-builder/pull/216/files#diff-f0272d7117ea8e9aba2b98d41e66e1a9R273) (feel free to make this function `async`).

Scenario: If the removal of the dist directory fails because of for example permission issues or because a file is still in use, a user would notice right away. With the current implementation, the user needs to wait for the whole build process to finish before receiving the error and the build is aborted.

test/lib/builder/builder.js
RandomByte: I see, now I get it 👍

lib/builder/builder.js
tommyvinhlam: Good point.

Cleaning before the actual build runs means that the target folder is always cleared, even if the new build has failed. But I can live with that.

mauriciolauffer:
Replaced by #306

PR #243, opened 2019-04-21
[FIX] ApplicationFormatter: detect the namespace from component controller

petermuessig:
When the manifest.json contains a Maven placeholder the namespace cannot
be detected properly. As a fallback the namespace will now be derived
from the component controller.

## Pull Request Checklist


petermuessig:
Maybe the fix is not the most optimal one, since it doesn’t use AST parsing to find the Component namespace and a regex instead but this definitely works (and is IMO as reliable as the AST one compared to the code and runtime effort being necessary to do AST parsing)!

lib/types/application/ApplicationFormatter.js
```diff
@@ -20,33 +20,60 @@ class ApplicationFormatter extends AbstractFormatter {
...
 	/**
 	 * Reads the manifest
-  *
+	 *
 	 * @param {Object} project
 	 * @returns {Promise<Object>} resolves with the json object
 	 */
 	readManifest(project) {
-		return readFile(path.join(project.path, project.resources.pathMappings["/"], "manifest.json"))
+		return readFile(path.join(project.path, project.resources.pathMappings["/"], "manifest.json"), "utf-8")
```
petermuessig: file encoding was missing before and should always be added!

lib/types/application/ApplicationFormatter.js
```diff
@@ -20,33 +20,60 @@ class ApplicationFormatter extends AbstractFormatter {
...
 					log.warn(`No "sap.app" ID configuration found in manifest of project ${project.metadata.name}`);
 					return;
 				}
-				project.metadata.namespace = manifest["sap.app"].id.replace(/\./g, "/");
-			}).catch((err) => {
+				// check app id for placeholder and try to guess namespace from component controller
+				const parts = appId.match(/^\$\{(.*)\}$/);
+				if (parts) {
+					log.warn(`No valid "sap.app" ID found in manifest of project ${project.metadata.name}. Guessing from Component.js...`);
+					return this.readComponentController(project).then((code) => {
```
tommyvinhlam: Wouldn't it make sense to enhance the namespace info to metadata section of the ui5.yaml and read it from there instead? I have mixed feelings about reading it from the component controller 🤔

lib/types/application/ApplicationFormatter.js
```diff
@@ -20,33 +20,60 @@ class ApplicationFormatter extends AbstractFormatter {
...
+						// use regex to determine namespace from component controller
+						const namespaceParts = code.match(/\.(?:\s+)?extend(?:\s+)?\((?:\s+)?["'](.*)\.Component["'](?:\s+)?,/);
```
codeworrior: instead of `(?:\s+)?` use the shorter and more readable `\s*`

codeworrior: second argument to extend(name, ...) is optional (although it rarely ever is missing)

codeworrior: Overall: why is the namespace read from the Component.js, not from the pom.xml? If the manifest says, ${project.artifactId} is the name, using the name from the extend call is not more than a guess.

lib/types/application/ApplicationFormatter.js
codeworrior: @tommyvinhlam: I guess the overall idea is: if some information can be found in existing (ideally: mandatory) files, then read it from there and do not add another redundant configuration for it (which increases the risk of inconsistencies).

lib/types/application/ApplicationFormatter.js
petermuessig: Reading it from pom.xml was also something I had briefly in mind about but I skipped it due to the fact to align it to the runtime behavior to determine the name of the Component. In addition, I also found out that the template which identified the issue has no pom.xml aside of the manifest.json which is also another issue and must not be happen. So, identifying from the pom.xml should be the better option and the manifest.json may only have such placeholder if a Maven build is aside.

petermuessig:
Enhanced the commit to read the pom.xml and be able to replace the Maven placeholders with the dedicated entries or properties from the pom.xml.

lib/types/application/ApplicationFormatter.js
```diff
@@ -19,12 +26,33 @@ class ApplicationFormatter extends AbstractFormatter {
...
+			return this.readManifest(project).then((manifest) => {
+				// check for a proper sap.app/id in manifest.json to determine namespace
+				const appId = manifest["sap.app"] && manifest["sap.app"].id;
+				if (!appId) {
 					log.warn(`No "sap.app" ID configuration found in manifest of project ${project.metadata.name}`);
 					return;
 				}
-				project.metadata.namespace = manifest["sap.app"].id.replace(/\./g, "/");
+				// check app id for placeholder and try to read the placeholder from the pom.xml
+				const parts = appId.match(/^\$\{(.*)\}$/);
```
tommyvinhlam: What about the idea of checking **${project.artifactId}** explicitly and maybe throw error in case of any other placeholder?

petermuessig: No, there could be any other placeholder be used for that!

lib/types/application/ApplicationFormatter.js
```diff
@@ -19,12 +26,33 @@ class ApplicationFormatter extends AbstractFormatter {
...
+				if (parts) {
+					log.warn(`No valid "sap.app" ID found in manifest of project ${project.metadata.name}. Reading "${parts[1]}" from pom.xml...`);
+					return this.readPOM(project).then((pom) => {
+						if (pom.project && pom.project.properties && pom.project.properties[parts[1]]) {
+							return pom.project.properties[parts[1]];
+						} else {
+							let obj = pom;
+							parts[1].split(".").forEach((part) => {
+								obj = obj[part];
+								return !obj;
```
RandomByte: I think this return is not necessary. According to [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach):

> There is no way to stop or break a forEach() loop other than by throwing an exception. If you need such behavior, the forEach() method is the wrong tool.

lib/types/application/ApplicationFormatter.js
```diff
@@ -19,12 +26,33 @@ class ApplicationFormatter extends AbstractFormatter {
...
+							});
+							return obj;
+						}
+					});
+				}
+				return appId;
+			}).then((appId) => {
+				project.metadata.namespace = appId ? appId.replace(/\./g, "/") : undefined;
 			}).catch((err) => {
 				log.verbose(`No manifest found for project ${project.metadata.name}.`);
```
RandomByte: This catch needs to move to only catch errors caused by manifest read errors. Currently also catches errors from reading the POM

lib/types/application/ApplicationFormatter.js
```diff
@@ -19,12 +26,33 @@ class ApplicationFormatter extends AbstractFormatter {
...
+						if (pom.project && pom.project.properties && pom.project.properties[parts[1]]) {
+							return pom.project.properties[parts[1]];
+						} else {
+							let obj = pom;
+							parts[1].split(".").forEach((part) => {
+								obj = obj[part];
+								return !obj;
+							});
+							return obj;
+						}
```
RandomByte: Please add some error handling (like log.warn similar to the manifest check) in case the placeholder could not be resolved.

lib/types/application/ApplicationFormatter.js
```diff
@@ -5,6 +5,13 @@ const {promisify} = require("util");
 const readFile = promisify(fs.readFile);
 const AbstractFormatter = require("../AbstractFormatter");
 
+const xml2js = require("xml2js");
```
RandomByte: Please move the require and parser initialization to the readPOM function. There is no need to always execute this code.

lib/types/application/ApplicationFormatter.js
```diff
@@ -19,12 +26,33 @@ class ApplicationFormatter extends AbstractFormatter {
...
+				const appId = manifest["sap.app"] && manifest["sap.app"].id;
+				if (!appId) {
 					log.warn(`No "sap.app" ID configuration found in manifest of project ${project.metadata.name}`);
 					return;
 				}
-				project.metadata.namespace = manifest["sap.app"].id.replace(/\./g, "/");
+				// check app id for placeholder and try to read the placeholder from the pom.xml
+				const parts = appId.match(/^\$\{(.*)\}$/);
+				if (parts) {
+					log.warn(`No valid "sap.app" ID found in manifest of project ${project.metadata.name}. Reading "${parts[1]}" from pom.xml...`);
```
RandomByte: Is this really a warning? Do we want developers to remove such placeholders? I thought it is still valid for many developers to use them.

lib/types/application/ApplicationFormatter.js
RandomByte: In addition, a similar catch for the POM read is required. The reason for them is to suppress the errors since it is still valid for applications to not have a namespace.

lib/types/application/ApplicationFormatter.js
petermuessig: I did bookmark this in my brain - now it is! Next time you will not see the return - might have been done wrong in several other places... :-)

lib/types/application/ApplicationFormatter.js
petermuessig: Good point... Fixed!

lib/types/application/ApplicationFormatter.js
petermuessig: Ok, done!

lib/types/application/ApplicationFormatter.js
petermuessig: Always running the re-initialization of the parser is a bit more expensive - but I don't care! In general, I agree with the require to be done lazy...

lib/types/application/ApplicationFormatter.js
petermuessig: No. Changed to .verbose

petermuessig:
This time I created a follow up commit which will be squashed before the merge later on...

petermuessig:
Hmm, I am wondering why the build is failing! The JSDoc build works locally but somehow centrally it doesn't! I now squashed the change locally manually as this maybe affects the build... :-/

lib/types/application/ApplicationFormatter.js
RandomByte: Typically the ApplicationFormatter is executed only once anyways.

RandomByte:
Looks pretty good 👍 

@tommyvinhlam to approve.

tommyvinhlam:
Rebase required

PR #245, opened 2019-04-24
[INTERNAL] Set up CI with Azure Pipelines

RandomByte:
Switching from Travis to Azure Pipelines can give us quicker builds, while removing pressure from the already over capacity SAP-Org Travis CI.

Some nice addition are the integrated [Test Result](https://dev.azure.com/sap/opensource/_build/results?buildId=224&view=ms.vss-test-web.build-test-results-tab) (failing example) and [Code Coverage](https://dev.azure.com/sap/opensource/_build/results?buildId=225&view=codecoverage-tab) reports.

However, Coveralls does not yet officially support Azure Pipelines Therefore we still need to execute a minimal Travis build for Coveralls reporting.


azure-pipelines.yml
```diff
@@ -0,0 +1,50 @@
...
+pool:
+  vmImage: $(imageName)
+
+steps:
+- task: NodeTool@0
+  inputs:
+    versionSpec: $(node_version)
+  displayName: 'Install Node.js'
+
+- script: npm install
```
matz3: AFAIK Travis CI uses `npm ci` instead of `npm install`:
https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#npm-ci-support
https://docs.npmjs.com/cli/ci.html

RandomByte: Done

PR #255, opened 2019-05-22
[FIX] Detect library namespace automatically

RandomByte:
If no manifest.json is present, fallback to .library. If that is not present, fallback to location of library.js file.

Also refactor read operations of UI5 specific, formatter relvevant
files.


RandomByte:
CC: just fyi @Thodd, @petermuessig

lib/types/AbstractUi5Formatter.js
```diff
@@ -0,0 +1,124 @@
...
+const {promisify} = require("util");
+const readFile = promisify(fs.readFile);
+
+/**
+ * Base class for formatters that require access to some UI5 specific resources
+ * like manifest.json or pom.xml
+ *
+ * @abstract
+ */
+class AbstractUi5Formatter extends AbstractFormatter {
```
codeworrior: Seeing 'Ui5' hurts my eyes...

lib/types/AbstractUi5Formatter.js
```diff
@@ -0,0 +1,124 @@
...
+	/**
+	 * Reads the projects manifest.json
+	 *
+	 * @returns {Promise<Object>} Resolves with a JSON representation of the content
+	 */
+	async getManifest() {
+		if (this._pManifest) {
+			return this._pManifest;
+		}
+		const fsPath = path.join(this._project.path, this.getSourceBasePath(), "manifest.json");
```
matz3: This won't work for libraries. A glob is required to find the file in a library project.

lib/types/library/LibraryFormatter.js
```diff
@@ -1,75 +1,178 @@
...
+				appId = await this.resolveMavenPlaceholder(appId);
+			} catch (err) {
+				throw new Error(
+					`Failed to resolve namespace maven placeholder of project ` +
+					`${this._project.metadata.name}: ${err.message}`);
+			}
+		}
+
+		const namespace = appId.replace(/\./g, "/");
+		log.verbose(`Namespace of project ${this._project.metadata.name} is ${namespace}`);
```
matz3: Also check whether the library namespace matches with the file path of `manifest.json`, `.library` or `library.js`

lib/types/library/LibraryFormatter.js
```diff
@@ -1,75 +1,178 @@
...
-					}, (error) => {
-						throw new Error(`Failed to check .library file for project ${project.id}, ` +
-							`error: ${error.message}`);
-					});
+			} catch (err) {
+				throw new Error(`Namespace resolution from .library failed for project ` +
+					`${this._project.metadata.name}: ${err.message}`);
+			}
+		}
+
```
matz3: If no manifest.json and no .library are available, check for `library.js` and derive the library name from the file path (like we already do in https://github.com/SAP/ui5-builder/blob/92b92656831cd136b118dcd9ad92a7651e4a05e9/lib/tasks/bundlers/generateLibraryPreload.js#L208-L230)

lib/types/AbstractUi5Formatter.js
matz3: Mine too 🙄

lib/types/AbstractUi5Formatter.js
RandomByte: Done

lib/types/library/LibraryFormatter.js
RandomByte: Done

lib/types/library/LibraryFormatter.js
RandomByte: Done

RandomByte:
Reminder to myself: **Squash it**

lib/types/AbstractUi5Formatter.js
matz3: Hmm so what about this?

RandomByte: Later in some other change 😬

It's private API anyways

nlunets:
@codeworrior @RandomByte @matz3 this change broke building application with all dependencies.

The main reason being that you now expect a namespace to be provided for libraries, and you want to read that from the LibraryFormatter...
However the library formatter is never called (neither is the application formatter for that matter), and apparently it has never been called :)

So right now you cannot even build the testsuite with that, this is what's breaking ui5-typescript master branch as well @bd82 FYI

RandomByte:
@nlunets thanks for reporting. Could you please share some additional information?
- What is broken? Is there an error message? Are artefacts missing from the build result?
- What version of the UI5 CLI are you referring to?
- How can we reproduce the issue?

I just did successful builds (`ui5 build --all`) of the latest master of the [OpenUI5 Sample App](https://github.com/SAP/openui5-sample-app) and the [OpenUI5 Testsuite](https://github.com/SAP/openui5/tree/master/src/testsuite) using the UI5 CLI v1.5.3.

Note that there have been follow up changes to this one resolving some issues we identified after merging this PR:
- https://github.com/SAP/ui5-builder/pull/262
- https://github.com/SAP/ui5-builder/pull/263

However, they all got released with the same UI5 Builder v1.3.1 release:
https://github.com/SAP/ui5-builder/compare/v1.3.0...v1.3.1

nlunets:
- Clone openui5
- Run yarn
- go into src/testsuite and add the @ui5/cli tooling there, otherwise you don't have the proper instlalation since openui5 depends on specific version of ui5-server and ui5-project https://github.com/SAP/openui5/blob/master/package.json#L32-L33
- `yarn add @ui5/cli --dev`
- run `ui5 build jsdoc --all` and see those errors

```
info builder:builder library sap.ui.core Skipping some tasks due to missing library namespace information. Your project might be missing a manifest.json or .library file. Also see: https://github.com/SAP/ui5-builder#library
info builder:builder library sap.ui.core 🔨 (1/1) Running task buildThemes...
```
There should be some extra steps here

```
info builder:builder library sap.ui.core 🔨 (1/3) Running task generateJsdoc...
WARNING: The @private tag does not permit a value; the value will be ignored. File: ManagedObjectModel.js, line: 497
WARNING: The @private tag does not permit a value; the value will be ignored. File: ManagedObjectModel.js, line: 506
info builder:builder library sap.ui.core 🔨 (2/3) Running task executeJsdocSdkTransformation...
```

nlunets:
Before the namespace was computed from the name 

```
const namespace = project.metadata.name.replace(/\./g, "/");
``` 

And now it's just not coming from anywhere because we're never going into the library / application formatter.

RandomByte:
The yarn.lock file in the OpenUI5 repository references old versions of some UI5 Tooling modules that seemingly lead to this issue. We need to investigate why this happens exactly.

In the meantime, please install the UI5 CLI globally and remove the local installation. This should solve the issue.

nlunets:
yep it works with removing the yarn.lock file apparently, thanks :)
Still, i'm not sure the formatter is called ;)

RandomByte:
It is, but in the erroneous setup and old version of the UI5 Builder is used that does not have the new namespace detection implemented in this PR.

PR #261, opened 2019-06-03
[INTERNAL] Bump ava to 2.0.0

RandomByte:
Relevant changelogs:
https://github.com/avajs/ava/releases/tag/v1.0.1
https://github.com/avajs/ava/releases/tag/v2.0.0

Relevant changes:
- AVA now requires at least Node.js 8.9.4. We require only 8.5.
  However, since this is a devDependency I guess we can live with that.
- t.throws() does not accept promises anymore. Use t.throwsAsync()
- Test titles now must be unique

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,70 +5,31 @@
 	"requires": true,
 	"dependencies": {
 		"@ava/babel-plugin-throws-helper": {
-			"version": "2.0.0",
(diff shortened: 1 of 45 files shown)


RandomByte:
In the ResourcePool.js test, some trailing whitespaces in the test strings have been removed by my IDE, causing the string size to fall below the expected.

This happened before, therefore I removed the trailing whitespaces and adapted the expected string size.

PR #267, opened 2019-06-18
[FIX] generateManifestBundle: Only glob files from project namespace

RandomByte:
manifestBundler can only include resources from paths that are based on the projects namespace (see regex in manifestBundler ZIP creation).

For other paths the bundling fails because yazl can't add absolute paths.

Also, only resources located in the projects "/resources" should be relevant for manifest bundling.

The latter resovles https://github.com/UI5Lab/UI5Lab-library-simple/issues/11 where with UI5 CLI v1.5.1 the only problem left is, that generateManifestBundle tries to bundle manifest.json resources found in the libraries /test-resources.

### DIFF
--- a/lib/processors/bundlers/manifestBundler.js
+++ b/lib/processors/bundlers/manifestBundler.js
@@ -1,6 +1,7 @@
 const path = require("path");
 const yazl = require("yazl");
 const resourceFactory = require("@ui5/fs").resourceFactory;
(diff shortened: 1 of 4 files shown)


matz3:
Shouldn't we still add some error handling within the `manifestBundler` processor to prevent adding absolute paths to the zip?

PR #273, opened 2019-06-24
[FIX] manifestCreator: Only consider component files called Component.js

RandomByte:
In the past files called *Component.js where also considered as
Components during manifest creation.

### DIFF
--- a/lib/processors/manifestCreator.js
+++ b/lib/processors/manifestCreator.js
@@ -159,7 +159,7 @@ async function createManifest(libraryResource, libBundle, descriptorVersion, _in
 		function findComponentPaths() {
 			const result = [];
 			const prefix = libraryResource.getPath().slice(0, - ".library".length);
-			const components = libBundle.getResources(/(?:[^/]+\/)*Component\.js$/);
+			const components = libBundle.getResources(/^(?:[^/]*\/)*Component\.js$/);
 			components.forEach((comp) => {
 				const relativePath = comp.getPath().slice(prefix.length);
 				if ( relativePath.lastIndexOf("/") >= 0 ) {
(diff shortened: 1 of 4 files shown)


lib/processors/manifestCreator.js
```diff
@@ -159,7 +159,7 @@ async function createManifest(libraryResource, libBundle, descriptorVersion, _in
 		function findComponentPaths() {
 			const result = [];
 			const prefix = libraryResource.getPath().slice(0, - ".library".length);
-			const components = libBundle.getResources(/(?:[^/]+\/)*Component\.js$/);
+			const components = libBundle.getResources(/^(?:[^/]*\/)*Component\.js$/);
```
codeworrior: The anchor was for sure missing, but the remainder of the pattern looks strange to me now. It accepts empty path segments (which shouldn't exist in the wild) and as the bundler is for libraries, I also don't know why a /Component.js should exist. It would have no namespace, which would be an error for a manifest.json.

nlunets: @codeworrior this code actually detect all components and will make the manifest creator complain if there is both a library and a Component.js.

Right now this create the following errors:
```
ERR! builder:processors:manifestCreator Package /resources/sap/fe/AppComponent.js contains both '*.library' and 'Component.js'. This is not supported by manifests, therefore the component won't be listed in the library's manifest.
```

`/^(?:[^/]*\/)*Component\.js$/` matches things like
- Component.js
- /Component.js
- /toto/Component.js
- /toto/roro/Component.js

Which actually is an issue because we for instance ship components internally that can be reused as part of the library...
What would be the correct behavior here ?

nlunets: Actually after looking at this a bit more and discussing with @codeworrior we agreed that the right regexp would . be
`^/(?:[^/]+\/)*Component\.js$` because it will find the right component files without matching an empty path in . the middle like the other one.
And since all resources path start with a slash that's coherent

PR #285, opened 2019-07-08
[FEATURE] Add new theme-library type

RandomByte:
Theme libraries do not require as many features as normal libraries.
For example they do not need a namespace. Also many build tasks are
not required.

### DIFF
--- a/lib/types/themeLibrary/ThemeLibraryFormatter.js
+++ b/lib/types/themeLibrary/ThemeLibraryFormatter.js
@@ -1,0 +1,86 @@
+const log = require("@ui5/logger").getLogger("types:themeLibrary:ThemeLibraryFormatter");
+const path = require("path");
+const AbstractUi5Formatter = require("../AbstractUi5Formatter");
+
+
+class ThemeLibraryFormatter extends AbstractUi5Formatter {
+	/**
+	 * Formats and validates the project
+  	 *
+	 * @returns {Promise}
+	 */
+	async format() {
+		const project = this._project;
+		await this.validate();
+
+		log.verbose("Formatting theme-library project %s...", project.metadata.name);
+		project.resources.pathMappings = {
+			"/resources/": project.resources.configuration.paths.src
+		};
+
+		if (project.resources.configuration.paths.test) {
+			// Directory 'test' is somewhat optional for theme-libraries
+			project.resources.pathMappings["/test-resources/"] = project.resources.configuration.paths.test;
+		} else {
+			log.verbose(`Ignoring 'test' directory for project ${project.metadata.name}.` +
+				"Either no setting was provided or the path not found.");
+		}
+	}
+
+	/**
+	 * Validates the project
+	 *
+	 * @returns {Promise} resolves if successfully validated
+	 * @throws {Error} if validation fails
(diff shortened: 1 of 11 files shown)


RandomByte:
TODO: Add tests

RandomByte:
Moved the previous changes in LibraryFormatter to https://github.com/SAP/ui5-builder/pull/392

RandomByte:
As discussed, release this together with a new specification version "1.1" to give developers a good error message when consuming updated OpenUI5 theme libraries.

For this, the theme-library formatter needs to be enhanced to check for spec version 1.1 or higher. Similar to what has been done in proxy PoC: https://github.com/SAP/ui5-project/pull/209/files#diff-bf8943f684859f68f7e7a4d3cfed68ddR275

lib/types/themeLibrary/ThemeLibraryFormatter.js
```diff
@@ -0,0 +1,86 @@
...
+	 *
+	 * @returns {Promise} resolves if successfully validated
+	 * @throws {Error} if validation fails
+	 */
+	validate() {
+		const project = this._project;
+		return Promise.resolve().then(() => {
+			if (!project) {
+				throw new Error("Project is undefined");
+			} else if (!project.metadata || !project.metadata.name) {
```
matz3: specVersion should be checked here

matz3:
TODO: Update relevant documentation

lib/types/themeLibrary/ThemeLibraryFormatter.js
```diff
@@ -0,0 +1,88 @@
...
+	 *
+	 * @returns {Promise} resolves if successfully validated
+	 * @throws {Error} if validation fails
+	 */
+	validate() {
+		const project = this._project;
+		return Promise.resolve().then(() => {
+			if (!project) {
+				throw new Error("Project is undefined");
+			} else if (project.specVersion === "0.1" || project.specVersion === "1.0") {
```
matz3: @RandomByte do you think this simple check is okay? With this we rely on the general specVersion check within ui5-project to be executed, otherwise also other specVersions would be allowed.

RandomByte: LGTM for the time being

lib/types/themeLibrary/themeLibraryType.js
```diff
@@ -0,0 +1,15 @@
...
+module.exports = {
+	format: function(project) {
+		return new ThemeLibraryFormatter({project}).format();
+	},
+	build: function({resourceCollections, tasks, project, parentLogger, buildContext}) {
+		return new ThemeLibraryBuilder({resourceCollections, project, parentLogger, buildContext}).build(tasks);
+	},
+
+	// Export type classes for extensibility
+	Builder: ThemeLibraryBuilder,
```
codeworrior: does "for extensibility" only mean that consumers can extend those classes or does it also mean that they can inject own implementations here via these two names? If so, the new expressions above should use the properties of the export, not the import names.

RandomByte: Yes, subclassing

PR #293, opened 2019-07-16
[FEATURE] Properties File Escaping

tobiasso85:
Add processor which escapes non ASCII characters in resources.

As per RFC 0007: [SAP/ui5-tooling#168](https://github.com/SAP/ui5-tooling/pull/168)


lib/processors/stringEscaper.js
```diff
@@ -0,0 +1,59 @@
+const escapeUnicode = require("escape-unicode");
+
+/**
+ * @see https://ascii.cl/
+ * @type {number}
+ */
+const NUMBER_OF_ASCII_CHARACTERS = 127;
```
Thodd: 128
rename to something like CHAR_CODE_OF_LAST_ASCII_CHARACTER ?

tobiasso85: done

lib/processors/stringEscaper.js
```diff
@@ -0,0 +1,61 @@
...
+	}
+	return {
+		modified,
+		string: result
+	};
+};
+
+module.exports = function({resources}) {
+	return Promise.all(resources.map((resource) => {
+		return resource.getString().then((resourceString) => {
```
matz3: `getString` always reads with "utf-8" encoding. This should be dependent on the encoding configured in the ui5.yaml file.

lib/tasks/escapeNonAsciiCharacters.js
```diff
@@ -0,0 +1,43 @@
...
+ * @param {string} parameters.options.pattern Pattern to locate the files to be processed
+ * @param {string} [parameters.options.sourceEncoding] source file encoding: "UTF-8" or "ISO-8859-1". Defaults to "ISO-8859-1"
+ * @returns {Promise<undefined>} Promise resolving with <code>undefined</code> once data has been written
+ */
+module.exports = function({workspace, options}) {
+	if (options.sourceEncoding && !Object.values(ENCODING_TYPES).includes(options.sourceEncoding)) {
+		return Promise.reject(new Error(`Invalid encoding specified: '${options.sourceEncoding}'. Must be one of ${Object.values(ENCODING_TYPES)}`));
+	}
+	if (options.sourceEncoding && options.sourceEncoding === ENCODING_TYPES.UTF_8) {
+		return Promise.resolve();
```
matz3: Afaik we must do the escaping for both encodings. The encoding should just specify how the file content is read.
Only this way we can make sure that it works in all environments, regardless of the encoding specified by servers.

Thodd: Done

lib/processors/stringEscaper.js
Thodd: Done

lib/processors/stringEscaper.js
```diff
@@ -0,0 +1,61 @@
+const escapeUnicode = require("escape-unicode");
+
+/**
+ * @see https://ascii.cl/
```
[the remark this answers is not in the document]
Thodd: Done

Thodd:
@tobiasso85:
Sorry, but @matz3 and I broke some tests. 
#notsorry

tobiasso85:
> @tobiasso85:
> Sorry, but @matz3 and I broke some tests.
> #notsorry

#fixedit #unbreakable

lib/lbt/bundle/AutoSplitter.js
```diff
@@ -226,7 +228,20 @@ class AutoSplitter {
 				props.store(out, "");
 				return out.toString().length();
 				*/
-				return fileContent.toString("latin1").length;
+
+				// input encoding is read from project configuration, default is ISO-8859-1
+				const encoding = resource.getProject() && resource.getProject().resources.propertiesFileEncoding;
```
RandomByte: I'm wondering, why is this necessary?

If the `escapeNonAsciiCharacters` tasks was executed, all .properties resources should be UTF-8 compatible by now.

Or are these resources coming from dependencies and not the project itself?

lib/processors/stringEscaper.js
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Escapes non ASCII characters with unicode escape sequences.
+ *
+ * @public
+ * @alias module:@ui5/builder.processors.stringEscaper
+ * @param {Object} parameters Parameters
+ * @param {module:@ui5/fs.Resource[]} parameters.resources List of resources to be processed
+ * @param {Object} [parameters.options] Options
+ * @param {string} [parameters.options.encoding="utf8"] resource file encoding based on supported
```
RandomByte: We need to agree on a place to make the transformation between ui5.yaml configuration (`UTF-8`/`ISO-8859-1`) and the Node.js encodings strings (`utf8`/`latin1`).

Currently I think I can put `"UTF-8"` in my ui5.yaml and it will be passed to `.toString` here which only understands `"utf8"`.

tobiasso85: the only "unsupported" one is: ISO-8859-1.
https://encoding.spec.whatwg.org/#names-and-labels
UTF8 / utf8 / UTF-8 are all supported by node

RandomByte: https://encoding.spec.whatwg.org/#names-and-labels is not the correct resource for what is supported by Node.js. It is referred to as being what "Modern Web browsers follow", not Node.js.

I just tested the following to confirm this:
```
> Buffer.from("🐬", "unicode-1-1-utf-8").toString()
Thrown:
TypeError [ERR_UNKNOWN_ENCODING]: Unknown encoding: unicode-1-1-utf-8
```

The [Node.js documentation](https://nodejs.org/api/buffer.html#buffer_buffers_and_character_encodings) states a list of strings that is supported. `"UTF-8"` seems to work, but is not documented and supposedly only supported for legacy reasons. There are even shortcuts for the documented `utf8` in the code: https://github.com/nodejs/node/blob/master/lib/internal/util.js#L104

I would prefer an explicit mapping to what is stated in the Node.js API docs just to be safe.

RandomByte: As discussed with @tobiasso85:

Make the `parameters.options.encoding` input parameter of the **task** reflect the ui5.yaml configuration options. So either `"UTF-8"` or `"ISO-8859-1"`. Make the parameter mandatory with no default (there is no point in using the task if the encoding is unknown).

The **processors** `parameters.options.encoding` input parameter should accept all Node.js supported encodings as listed at https://nodejs.org/api/buffer.html#buffer_buffers_and_character_encodings. There should not be any additional mapping (like `ISO-8859-1` to `latin1`) in the processor.

tobiasso85: indeed

lib/lbt/bundle/AutoSplitter.js
tobiasso85: the latter

lib/processors/nonAsciiEscaper.js
```diff
@@ -0,0 +1,99 @@
...
+const encodingMap = {
+	"UTF-8": "utf8",
+	"ISO-8859-1": "latin1",
+};
+
+/**
+ * @param {string} encoding encoding labels: "UTF-8" and "ISO-8859-1"
+ * @returns {string} node.js character encoding string, e.g. utf8 and latin1
+ */
+module.exports.getEncodingFromNiceName = function(encoding) {
```
Thodd: Doesn't this function fit much better in the task?
See @RandomByte's earlier comment.

It's never called in the processor itself.
I think it is cleaner to keep the processor solely node-based, as it only understands the node char encodings anyway.

This fits much better into the task, which relies on external encodings like used by UI5.
It's also only called by the task so...

Also:
Maybe call it getEncodingNameFromAlias() ?

Thodd:
Also, @tobiasso85's earlier commits had a validation for the allowed encoding, which I removed with @matz3.
IIRC we thought the validation the accepted encoding names should be done in the `ui5-project` upon parsing the projects config.

lib/processors/nonAsciiEscaper.js
tobiasso85: @Thodd **the place**: if this method `getEncodingFromNiceName` is within the processor it can be easier consumed from the server and doesn't require a task.
@Thodd **the name**: getEncodingNameFromAlias is also good. I like it even better.

tobiasso85:
> Also, @tobiasso85's earlier commits had a validation for the allowed encoding, which I removed with @matz3.
> IIRC we thought the validation the accepted encoding names should be done in the `ui5-project` upon parsing the projects config.

But the project- and library-formatter set the default, so the task doing the validation is fine imho.

RandomByte:
> IIRC we thought the validation the accepted encoding names should be done in the ui5-project upon parsing the projects config.

The validation of the ui5.yaml configuration should happen in ui5-project, yes. Additionally, the tasks API should validate its parameters imho. It might be used as a standalone module, without the ui5-project.

lib/processors/nonAsciiEscaper.js
RandomByte: @matz3 @tobiasso85 and myself talked about this yesterday. We came to the agreement that while it fits better in the task (because it bridges between the UI5- and node-native world), its something you need when calling the processor. So it's more convenient for the UI5 Server and the bundler tasks to have it available there.

lib/tasks/escapeNonAsciiCharacters.js
```diff
@@ -0,0 +1,30 @@
...
+ * @param {Object} parameters Parameters
+ * @param {module:@ui5/fs.DuplexCollection} parameters.workspace DuplexCollection to read and write files
+ * @param {Object} parameters.options Options
+ * @param {string} parameters.options.pattern Glob pattern to locate the files to be processed
+ * @param {string} parameters.options.encoding source file encoding either "UTF-8" or "ISO-8859-1"
+ * @returns {Promise<undefined>} Promise resolving with <code>undefined</code> once data has been written
+ */
+module.exports = async function({workspace, options}) {
+	if (!options.encoding) {
+		throw new Error("No Encoding specified");
```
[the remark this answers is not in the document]
tobiasso85: done
(added prefix and option to the error message with f0be15c)

test/lib/builder/builder.js
```diff
@@ -84,7 +84,7 @@ test("Build application.a", (t) => {
 	return builder.build({
 		tree: applicationATree,
 		destPath,
-		excludedTasks: ["generateComponentPreload", "generateStandaloneAppBundle", "generateVersionInfo"]
+		excludedTasks: ["escapeNonAsciiCharacters", "generateComponentPreload", "generateStandaloneAppBundle", "generateVersionInfo"]
```
RandomByte: Instead of excluding the task everywhere, please rather add the necessary `"propertiesFileEncoding": "ISO-8859-1"` configuration to the project trees where necessary.

This is a consequence of our decision to make the defaulting in the type-formatters instead of the -builders. Others, that use our `build()` API directly, need to do the same (and not exclude the task).

test/lib/lbt/bundle/AutoSplitter.js
```diff
@@ -16,7 +16,14 @@ function createMockPool(dependencies) {
 			if (name === "c.js") {
 				info.compressedSize = 512;
 			}
-			return {info, buffer: async () => name.padStart(2048, "*")};
+			return {
+				info,
+				buffer: async () => Buffer.from(name.padStart(2048, "*")),
+				getProject: () => undefined,
```
RandomByte: Because of this, the newly added code in AutoSplitter is **never** tested:
https://coveralls.io/builds/24765632/source?filename=lib%2Flbt%2Fbundle%2FAutoSplitter.js#L240

Please test it. Why not mock the project configuration here?

test/lib/processors/nonAsciiEscaper.js
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Executes string escaping. Returns <code>undefined</code> if nothing was escaped.
+ *
+ * @param {string} input string
+ * @param {Object} [options]
+ * @returns {Promise<string|undefined>} escaped string if non-ascii characters present, <code>undefined</code> otherwise
+ */
+const escape = async function(input, options) {
+	let result = undefined;
+	const resource = {
```
RandomByte: Why don't you just create a real Resource? https://sap.github.io/ui5-tooling/api/module-@ui5_fs.Resource.html

It's not clear to me what became easier with this approach.

test/lib/processors/nonAsciiEscaper.js
```diff
@@ -0,0 +1,101 @@
+const test = require("ava");
```
RandomByte: I'm missing a test for `latin1` encoded strings

test/lib/processors/nonAsciiEscaper.js
```diff
@@ -0,0 +1,101 @@
...
+
+test("getEncodingFromAlias", (t) => {
+	t.is("utf8", nonAsciiEscaper.getEncodingFromAlias("UTF-8"));
+});
+
+test("getEncodingFromAlias invalid", (t) => {
+	const error = t.throws(function() {
+		nonAsciiEscaper.getEncodingFromAlias("asd");
+	});
+	t.is(error.message, `Encoding "asd" is not supported. Only UTF-8,ISO-8859-1 are allowed values`);
```
RandomByte: > Only UTF-8,ISO-8859-1

Can be get a space between them? 🙂

test/lib/builder/builder.js
tobiasso85: done

test/expected/build/application.b/dest/manifest-bundle/i18n/i18n.properties
```diff
@@ -1 +1,2 @@
-title=app-i18n
\ No newline at end of file
+title=app-i18n
+fame=Stra�e
```
RandomByte: Why is this not escaped?

test/lib/lbt/bundle/AutoSplitter.js
tobiasso85: done

test/lib/processors/nonAsciiEscaper.js
tobiasso85: done

test/lib/processors/nonAsciiEscaper.js
tobiasso85: done

test/lib/processors/nonAsciiEscaper.js
tobiasso85: done

test/expected/build/application.b/dest/manifest-bundle/i18n/i18n.properties
tobiasso85: config option was not specified

lib/lbt/bundle/AutoSplitter.js
```diff
@@ -226,7 +228,27 @@ class AutoSplitter {
...
+						resources: [resource.resource],
+						options: {
+							encoding
+						}
+					});
+				}
+
+				const fileContent = await resource.buffer();
+
+				return fileContent.toString().length;
```
matz3: Right now this will not do escaping if there is no project (e.g. when bundler is used with manually created FS adapters). And in this case the behavior changes to read the files with utf-8 encoding, not latin1.

I would prefer to also do the escaping in that scenario and use latin1, as that has been the behavior so far.

lib/lbt/bundle/Builder.js
```diff
@@ -442,9 +444,26 @@ class BundleBuilder {
...
+					options: {
+						encoding
+					}
+				});
+			}
+
 			const fileContent = await resource.buffer();
-			outW.write( makeStringLiteral( fileContent.toString("latin1") ) );
+
+			outW.write( makeStringLiteral( fileContent.toString() ) );
```
matz3: Same as in AutoSplitter

test/lib/lbt/bundle/AutoSplitter.js
```diff
@@ -16,7 +16,22 @@ function createMockPool(dependencies) {
...
+					return {
+						"resources": {
+							"configuration": {
+								"propertiesFileEncoding": "ISO-8859-1"
+							}
+						}
+					};
+				},
+				resource: {
+					getBuffer: () => Buffer.from(name.padStart(2048, "*"))
```
matz3: Shouldn't this be an `async` function?

test/lib/lbt/bundle/AutoSplitter.js
```diff
@@ -183,20 +200,37 @@ test.serial("_calcMinSize: uglify js resource", async (t) => {
...
 		findResourceWithInfo: function() {
+			let content = "1234ß";
 			return {
-				buffer: async () => "1234"
+				buffer: async () => Buffer.from(content),
+				resource: {
+					setString: (string) => {
+						content = string;
+					},
+					getBuffer: () => Buffer.from(content, "latin1")
```
matz3: `async` function?

test/lib/processors/nonAsciiEscaper.js
```diff
@@ -0,0 +1,104 @@
...
+	});
+	return nonAsciiEscaper({resources: [resource], options});
+};
+
+test("Replace ISO-8859-1 characters", async (t) => {
+	t.plan(1);
+
+	const input = `Viele Grüße`;
+	const expected = `Viele Gr\\ufffd\\ufffde`;
+	const [resource] = await escape(input, {}, "latin1");
```
matz3: The default encoding of the nonAsciiEscaper is utf8, right?
The expected string is not correct.

lib/lbt/bundle/AutoSplitter.js
Thodd: done, moved to separate module

lib/lbt/bundle/Builder.js
Thodd: done, moved to separate module

test/lib/lbt/bundle/AutoSplitter.js
Thodd: yes you're right

test/lib/lbt/bundle/AutoSplitter.js
Thodd: done

test/lib/processors/nonAsciiEscaper.js
Thodd: done.

Thodd: add additional tests too

test/expected/build/application.b/dest/manifest-bundle/i18n/i18n.properties
tobiasso85: done

test/expected/build/application.b/standalone/resources/sap-ui-custom.js
```diff
@@ -5,10 +5,10 @@ jQuery.sap.registerPreloadedModules({
 	"id1/embedded/i18n/i18n_de.properties":'title=embedded-i18n_de',
 	"id1/embedded/i18n_fr.properties":'title=embedded-i18n_fr-wrong',
 	"id1/embedded/manifest.json":'{"_version":"1.1.0","sap.app":{"_version":"1.1.0","id":"id1.embedded","type":"component","applicationVersion":{"version":"1.2.2"},"embeddedBy":"../","title":"{{title}}"}}',
-	"id1/i18n.properties":'title=app-i18n-wrong',
-	"id1/i18n/i18n.properties":'title=app-i18n',
-	"id1/i18n/i18n_de.properties":'title=app-i18n_de',
-	"id1/i18n/l10n.properties":'title=app-i18n-wrong',
+	"id1/i18n.properties":'title=app-i18n-wrong\nfame=Stra\\u00dfe',
```
matz3: This test still seems strange to me. Aren't the input files encoded with different encodings? (i18n.properties vs. i18n/i18n.properties)
How would this work here?

lib/lbt/utils/escapePropertiesFiles.js
```diff
@@ -0,0 +1,29 @@
...
+	await nonAsciiEscaper({
+		resources: [resource.resource],
+		options: {
+			encoding
+		}
+	});
+
+	const fileContent = await resource.buffer();
+
+	return fileContent.toString(encoding);
```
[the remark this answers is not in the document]
tobiasso85: done

lib/lbt/bundle/AutoSplitter.js
```diff
@@ -226,7 +227,12 @@ class AutoSplitter {
 				props.store(out, "");
 				return out.toString().length();
 				*/
-				return fileContent.toString("latin1").length;
+
+				// Since AutoSplitter is also used when splitting non-project resources (e.g. dependencies)
+				// *.properties files should be escaped if encoding option is specified
+				const fileContent = await escapePropertiesFiles(resource);
```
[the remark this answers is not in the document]
tobiasso85: done

test/expected/build/application.b/standalone/resources/sap-ui-custom.js
tobiasso85: it uses the builder
via writePreloadFunction -> writePreloadModule
and reads them as latin1 and then perform the escaping in
lib/lbt/bundle/Builder.js line 448
so this is valid imho

test/lib/builder/builder.js
```diff
@@ -342,7 +342,8 @@ test("Build theme.j even without an library", (t) => {
 	const expectedPath = "./test/expected/build/theme.j/dest";
 	return builder.build({
 		tree: themeJTree,
-		destPath
+		destPath,
+		excludedTasks: ["escapeNonAsciiCharacters"]
```
RandomByte: Why is the task excluded here? I would suspect the theme to not have any .properties files anyways

test/lib/types/application/ApplicationFormatter.js
```diff
@@ -96,6 +96,16 @@ test("validate: empty resources", async (t) => {
...
+test("validate: test invalid encoding", async (t) => {
+	const myProject = clone(applicationBTree);
+	myProject.resources.configuration.propertiesFileSourceEncoding = "test";
+	const applicationFormatter = new ApplicationFormatter({project: myProject});
+
+	const error = await t.throwsAsync(applicationFormatter.validate(myProject));
+	t.is(error.message, `Invalid properties file encoding specified for project application.b: encoding provided: test. Must be either "ISO-8859-1" or "UTF-8".`,
+		"Missing source directory caused error");
+});
+
```
RandomByte: The only thing I'm missing here is a test for the defaulting. An explicit test that a project without a `propertiesFileSourceEncoding` configuration, is being defaulted to having a `propertiesFileSourceEncoding: "ISO-8859-1"` configuration.

For the LibraryFormatter this assert has been added to the test `validate: test directory does not exist`. But a separate test would be even better.

test/lib/builder/builder.js
tobiasso85: done

test/lib/types/application/ApplicationFormatter.js
tobiasso85: done

tobiasso85: fixed with: 30ada8b

test/lib/builder/builder.js
tobiasso85: fixed with: 30ada8b

RandomByte:
Fine for me. **Please use "Squash and merge"!**

PR #294, opened 2019-07-16
[FIX] versionInfo: Use correct buildTimestamp format

matz3:


RandomByte:
Should `buildTimestamp` be UTC or in the systems time zone 🤔

matz3:
It is the system time zone in the Maven build, that's why I did it the same here.
I think UTC would be much better, but we should then align it with the Maven build.

cc @petermuessig @codeworrior

PR #295, opened 2019-07-18
[FEATURE] Theme Build: Add compress option to minify output

matz3:


matz3:
This now by default activates the compress option when running a build, which is the expected behavior I think.
The server will still serve the "debug" CSS.

matz3:
Done

PR #299, opened 2019-07-23
[INTERNAL] coverage: exclude .eslintrc.js

tobiasso85:
Exclude eslint config file since it is not treated as source code which
should be covered by a test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
 			"text-summary"
 		],
 		"exclude": [
+			".eslintrc.js",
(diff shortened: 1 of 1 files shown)


RandomByte:
Somehow all other UI5 Tooling projects already contain this exclude. I have no clue why it is missing in this project 🙄

Example: https://github.com/SAP/ui5-project/blob/master/package.json#L66

matz3:
I just came across this as well... 😂

PR #318, opened 2019-08-27
[FIX] Add 'sap.ui.fl' dependency to manifest.json

dstork:
If changes are present during runtime, the 'sap.ui.fl' dependency should be added to the manifest.json so the changes contained in the changes bundled can be applied to the application


lib/tasks/bundlers/generateFlexChangesBundle.js
```diff
@@ -20,6 +24,42 @@ module.exports = function({workspace, options}) {
 		pathPrefix = `/resources/${options.namespace}`;
 	}
 
+	function updateJson(data) {
+		const mLibs = data["sap.ui5"].dependencies.libs;
```
codeworrior: can't it happen that there's no dependencies/libs section yet in the manifest?

Lonwyr: at least if createManifest > createSapUi5 > create  > depdenencies is running / the creator is doing everything right, it should be fine.
In general the dependencies is flagged as mandatory while I only found a "have to be filled ... for ui5 apps" for the libs section.

Lonwyr: safety first... now within the mandatory part we always have the section not explicitly mentioned as mandatory

matz3: As most mandatory sections are not checked at runtime, I would expect the tooling to do the same. It should not expect that the given manifest is valid w.r.t. the schema.

lib/tasks/bundlers/generateFlexChangesBundle.js
```diff
@@ -32,8 +74,9 @@ module.exports = function({workspace, options}) {
 			});
 		})
 		.then((processedResources) => {
-			return Promise.all(processedResources.map((resource) => {
+			return Promise.all(processedResources.map(async (resource) => {
 				log.verbose("Writing flexibility changes bundle");
+				await updateFLdependency();
```
matz3: Processors return an array of resources, which should all be written to the workspace, although (currently) in this case it always returns one resource.

But updating the manifest should only happen once, right? So I would expect it to be done after the `Promise.all`.

test/expected/build/application.i/dest/manifest.json
```diff
@@ -1,13 +1 @@
...
-        "id": "application.i",
-        "type": "application",
-        "applicationVersion": {
-            "version": "1.2.2"
-        },
-        "embeds": ["embedded"],
-        "title": "{{title}}"
-    }
-}
+{"_version":"1.1.0","sap.app":{"_version":"1.1.0","id":"application.i","type":"application","applicationVersion":{"version":"1.2.2"},"embeds":["embedded"],"title":"{{title}}"},"sap.ui5":{"dependencies":{"libs":{"sap.ui.layout":{},"sap.ui.core":{},"sap.m":{},"sap.ui.fl":{}}}}}
```
matz3: I don't think that this is expected. When modifying the manifest IMO, it should still be formatted, preferably with the same indentation as before, but I think tabs should be fine.

matz3:
See https://github.com/SAP/ui5-builder/pull/319

PR #330, opened 2019-09-09
[INTERNAL] In-range update of npm dependencies + fix terser tests

RandomByte:
See https://github.com/terser/terser/blob/master/CHANGELOG.md#v430

> Functions passed to other functions in calls are now wrapped in
> parentheses by default, which speeds up loading most modules

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,18 +43,18 @@
 			}
 		},
 		"@babel/core": {
-			"version": "7.5.5",
-			"resolved": "https://registry.npmjs.org/@babel/core/-/core-7.5.5.tgz",
-			"integrity": "sha512-i4qoSr2KTtce0DmkuuQBV4AuQgGPUcPXMr9L5MyYAtk06z068lQ10a4O009fe5OB/DfNV+h+qqT7ddNV8UnRjg==",
+			"version": "7.6.0",
+			"resolved": "https://registry.npmjs.org/@babel/core/-/core-7.6.0.tgz",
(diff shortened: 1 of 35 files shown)


codeworrior:
Instead of adapting the tests, we IMO better should set the new option [`wrap_func_args`](https://github.com/terser/terser) to `false`.

RandomByte:
Sorry, I put that information in the second commit and forgot to edit the PR info.

Terser added a new `wrap_func_args` parameter which defaults to true.
PR implementing it: https://github.com/terser/terser/pull/433
Documentation: https://github.com/terser/terser#output-options => `wrap_func_args` (which refers to https://github.com/nolanlawson/optimize-js)

codeworrior:
Okay, only now I read your initial comment, so you have been aware of the change.

In my opinion, the opposite is true for UI5. Our library preloads rely on lazy parsing. Only when lazy parsing is used, loading the huge preloads isn't too bad as the browser only parses and compiles what's needed. 

But when we force eager parsing by adding the parentheses, all functions will be parsed and compiled after loading a library-preload. Measurements in the past have shown that this makes a very significant difference for the startup times of UI5 apps.

codeworrior:
@pskelin can you confirm that eager parsing would be a severe drawback for UI5 library preloads, after all we know?

RandomByte:
Alternative PR, setting tersers `wrap_func_args` option to false: https://github.com/SAP/ui5-builder/pull/331

Issue #338, opened 2019-09-20
Preload bundles do not support modules defining global variables


kristian:
In case a module is defining a variables in global scope, such as (only a selection):

```
sap/ui/thirdparty/blanket.js
sap/ui/thirdparty/flexie.js
sap/ui/thirdparty/hyphenopoly/hyphenEngine.asm.js
sap/ui/thirdparty/mobify-carousel.js
sap/ui/thirdparty/require.js
sap/ui/thirdparty/sinon-ie.js
sap/ui/thirdparty/sinon-server.js
sap/ui/thirdparty/swipe-view.js
sap/ui/thirdparty/zyngascroll.js
```

Building a preload bundle will result in the variables not be defined on global scope any longer, as the preload generation will wrap the module into a function itself, thus rendering the globally defined vars as function scoped vars instead.

## Expected Behavior

Building a preload bundle with a depdendency to e.g. sap/ui/thirdparty/require should still expose requirejs, require and define on global scope, just like loading the application w/o a preload bundle.

## Current Behavior

Global variables are not exposed, this could lead to side effects with other modules.

## Context
- UI5 Module Version: `1.5.1`

## Affected components *(if known)*

- [X] [ui5-builder](https://github.com/SAP/ui5-builder)
- [ ] [ui5-server](https://github.com/SAP/ui5-server)
- [ ] [ui5-cli](https://github.com/SAP/ui5-cli)
- [ ] [ui5-fs](https://github.com/SAP/ui5-fs)
- [ ] [ui5-project](https://github.com/SAP/ui5-project)
- [ ] [ui5-logger](https://github.com/SAP/ui5-logger)

codeworrior:
After looking deeper into it, this is a bit different than originally expected. The builder already _can_ handle modules with global vars correctly and it contains code to identify them.

But in case raw module info is maintained in `.library` files, this info supersedes the analysis (which is even skipped then). On the other side, raw-module info was never meant to be complete, it only should enhance the analysis results. 

The raw module info has been given priority over the analysis step when support for standard AMD APIs was added (_I can't reference the corresponding commit as this was implemented in an internal repo before going open source. But the code is [here](https://github.com/SAP/ui5-builder/blob/master/lib/lbt/resources/ResourcePool.js#L64-L71)_). The analysis of standard AMD bundles caused issues when the embedded modules are not available to the UI5 tooling. This approach should be revised so that the raw-module info and the analysis results can be  merged again. This will fix the issue reported here.

[...]

PR #340, opened 2019-09-22
[FIX] Bundling: merge dependency analysis results with raw module infos

codeworrior:
Fixes https://github.com/SAP/ui5-builder/issues/338

### DIFF
--- a/lib/lbt/bundle/Resolver.js
+++ b/lib/lbt/bundle/Resolver.js
@@ -61,13 +61,16 @@ class BundleResolver {
 						|| resource.info == null
 						|| resource.info.subModules.length === 0
 						|| /(?:^|\/)library.js$/.test(resource.info.name) ) {
-					return {resource, decomposable:false};
+					return {resource, decomposable: false};
 				}
 
-				// it might look more natural to expect 'all' embedded modules to exist in the pool,
-				// but expecting only one module is a more conservative approach
-				return Promise.all(resource.info.subModules.map((sub) => pool.findResource(sub).catch(() => false)))
-					.then((modules) => ({resource, decomposable: modules.some(($) => ($))}));
+				return Promise.all(
+					resource.info.subModules.map((sub) => pool.findResource(sub).catch(() => false))
+				).then((modules) => {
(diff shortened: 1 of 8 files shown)


codeworrior:
Note reg. the 2nd commit 265e5b64ea521993628e311ca3f27b6985692189: 
Due to the wrong handling of `findResource's` result, all bundles have been rated as 'decomposable' and the then needed but missing embedded resources caused 'uncaught promise rejection' errors.

Note to self: According to its name, `findResource` should not fail with an error but return null. Another `getResource` method could be introduced with the current behavior of `findResource` - if needed.

lib/lbt/bundle/Resolver.js
```diff
@@ -56,8 +56,16 @@ class BundleResolver {
...
-				return moduleInfo && moduleInfo.subModules.length > 0 && !/(?:^|\/)library.js$/.test(moduleInfo.name);
+			function isDecomposableBundle(moduleInfo) {
+				if ( moduleInfo == null
+						|| moduleInfo.subModules.length === 0
+						|| /(?:^|\/)library.js$/.test(moduleInfo.name) ) {
+					return false;
+				}
+				// it might look more natural to expect 'all' embedded modules to exist in the pool,
+				// but expecting only one module is a more conservative approach
+				return Promise.all(moduleInfo.subModules.map((sub) => pool.findResource(sub).catch(() => false)))
```
matz3: As already [mentioned by @kristian](https://github.com/SAP/ui5-builder/commit/265e5b64ea521993628e311ca3f27b6985692189#r35226030) I think this won't work. The line is also not covered by any test.

codeworrior: Yes, you're both totally right.

codeworrior:
Note: this PR requires a merge of PR https://github.com/SAP/ui5-builder/pull/341 to make the unit test for bundles work. Without that PR, the content of bundle modules is not properly recognized.

codeworrior:
@kristian could you please test this PR again in your context? The new unit and integration tests should cover your scenario, too. But I might have missed something.

PR #341, opened 2019-09-26
[FIX] Improve recognition of main module in case of bundles

codeworrior:
This is a port of several changes from the Maven tooling, which have been made after the initial migration of the bundle tooling. Additionally, code coverage has been slightly improved by adding additional unit and integration tests. Errors that have been revealed by the new tests have been fixed.

Details about the changes:

----
Change 3312369 (improve recognition of main module in case of bundles):

So far, the module analyzer always assumed the first module definition to be the main definition. This failed at least for CVOM bundles but might also fail for others.

The analysis has been improved and now accepts a module as main module only in the following cases:
- if it is an unnamed module
- if it is a named module whose name matches the external name
- if it is a named module and there's only one module (no bundle)

If multiple module definitions match the above criteria, this is
reported as an error.

Additionally, AMD special dependencies ('require', 'exports', 'module') are no longer reported in the dependency info.

----
Change 3945349 (JSModuleAnalyzer fails on modules that don't declare a name)

Basically, the JavaScript implementation was less sensitive to this issue. But if a resource could have a dependency to another resource named "null", the JavaScript tooling would fail, too. Although this looked like a purely theoretical scenario, adding a protective 'if' was trivial enough to just do it. 

----
Change 3562114 (fix dep. analysis of new evo bundles)

- recognize and evaluate sap.ui.require.preload calls
- write and extract bundle name and raw module names

### DIFF
--- a/lib/lbt/analyzer/JSModuleAnalyzer.js
+++ b/lib/lbt/analyzer/JSModuleAnalyzer.js
@@ -470,7 +470,7 @@ class JSModuleAnalyzer {
 			let i = 0;
 
 			// get the documentation from a preceding comment
-			let desc = getDocumentation(defineCall);
+			const desc = getDocumentation(defineCall);
 
 			// determine the name of the module
 			let name = null;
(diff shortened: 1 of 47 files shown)


codeworrior:
I intended to squash them. That's why I already updated the initial comment with a 'squashed' commit message.

matz3:
👍

PR #346, opened 2019-10-08
[INTERNAL] Update JSDoc Template

matz3:
Update JSDoc generation template


tobiasso85:
tested with #296 once with the existing jsdoc version 3.5.5 and 3.6.3 with 2 libraries and there was no difference in the produced api.json files

RandomByte:
514ae77 has no prefix. Either add `[INTERNAL]` to it or squash and merge this PR as a `[FIX]`

PR #354, opened 2019-10-15
[INTERNAL] JSModuleAnalyzer: Remove invalid log usage

tobiasso85:
Change usage from non existing log.debug to log.verbose.

This code was not yet covered by tests. The new tests ensure that the code is run through and no error is thrown.


tobiasso85:
> I would rather fix the existing test. The log lines ARE already covered by the EVOBundle test in [JSModuleAnalyzer](https://github.com/SAP/ui5-builder/blob/master/test/lib/lbt/analyzer/JSModuleAnalyzer.js#L292-L320). But the test unfortunately totally stifles all errors.
> 
> The issue is caused by line https://github.com/SAP/ui5-builder/blob/master/test/lib/lbt/analyzer/JSModuleAnalyzer.js#L97,

Good point! Addressed it with 4ce5d9d

test/lib/lbt/analyzer/JSModuleAnalyzer.js
```diff
@@ -287,8 +287,6 @@ test.cb("OldStyleBundleV2", (t) => {
 	);
 });
 
-// Note: this test still fails as the evo-style bundles don't declare the names of the
-// contained raw-modules any longer.
```
RandomByte: Was this comment removed intentionally?

codeworrior: I guess, yes. Because this TODO was, what #341 wanted to addressed (migration of 3562114).

tobiasso85: I thought this fix made the tests not fail anymore therefore I removed it

PR #356, opened 2019-10-21
[INTERNAL] Update JSDoc to 3.6.3

RandomByte:
Re-open version range in package.json to patches like grunt-jsdoc is
doing:
https://github.com/krampstudio/grunt-jsdoc/blob/7b4c6eb9db6ae207b18f1fdf37bc71d8c98c76da/package.json#L44

Minor updates tend to break the UI5 JSDoc plugin.


RandomByte:
Well we already mention it here: https://github.com/SAP/ui5-builder/commit/78e2a229f2ae11ca37538a75ac6746ff92af7b84 😅

matz3:
Ah I've missed that 👍

PR #360, opened 2019-10-29
[INTERNAL] introduce clean css

tobiasso85:
Introduce separate CSS optimization processor such that it can be used standalone, e.g. when serving files in ui5 server and to have a better css compression


RandomByte:
Could you add some context for those who where not involved in the discussion? (like me 🥺)

Questions:  
I guess the CSS compression gets better through the use of [clean-css](https://github.com/jakubpawlowicz/clean-css)? And this is in addition to the less internal compression? Or should it replace it?

Is file size reduction the only motivation for this change?

What is the reason for for having two parameters, `compress` and `compressJSON`? Isn't this more of a "mode"? I.e. does it make sense to ever set both to `true`?

tobiasso85:
> Could you add some context for those who where not involved in the discussion? (like me 🥺)
> 
> Questions:
> I guess the CSS compression gets better through the use of [clean-css](https://github.com/jakubpawlowicz/clean-css)? And this is in addition to the less internal compression? Or should it replace it?
> 
> Is file size reduction the only motivation for this change?
> 
> What is the reason for for having two parameters, `compress` and `compressJSON`? Isn't this more of a "mode"? I.e. does it make sense to ever set both to `true`?

You're right, let's discuss this again with all. This is also more trying out this solution. It was taken into account because:
using less to compress is in newer versions deprecated: http://lesscss.org/usage/#less-options-compress-deprecated-
clean-css has more options and we can stay backward compatible with the color shortening: https://github.com/jakubpawlowicz/clean-css#compatibility-modes

RandomByte:
## Scenario: Building `sap.m` on [OpenUI5](https://github.com/SAP/openui5/commit/04903edb5cf792b8f8bb4d90beb3c378e09d9041)
**master**: Average of 19.007 seconds over 10 runs  
**clean-css (1f8f7ed0bc873891a9c90eb199861e901488a904)**: Average of 20.522 seconds over 10 runs  
**Difference**: 1.515 seconds

Command: `ui5 build`  
Benchmark tool: [hyperfine](https://github.com/sharkdp/hyperfine) with one warmup  
Hardware: MacBook Pro 2017 - i7 2,9GHz/16GB

RandomByte:
As discussed, closing this PR.

RandomByte:
As it turnes out, the less compression feature has serious bugs. See https://github.com/SAP/less-openui5/pull/112

Switching to clean-css might resolve those with little performance penalty.

RandomByte:
Rebased onto current master and resolved conflicts. There are still failing tests however which an expert should check for correctness.

lib/processors/themeBuilder.js
```diff
@@ -30,9 +30,12 @@ class ThemeBuilder {
 	 * @param {module:@ui5/fs.Resource[]} resources Library files
 	 * @param {Object} [options] Build options
 	 * @param {boolean} [options.compress=false] Compress build output (CSS / JSON)
+	 *						DEPRECATED: Use [cssOptimizer]{@link module:@ui5/builder.processors.cssOptimizer} to compress CSS output.
```
RandomByte: A warning log in the code would also be reasonable. Similar to https://github.com/SAP/ui5-fs/blob/master/lib/resourceFactory.js#L72

tobiasso85: done with 75c54b9

PR #380, opened 2019-12-11
[FEATURE] Add included/excludedDependencies parameter

kristian:
When setting the buildDependencies parameter to "true" builder
used to build all dependencies of the project. By adding two new
parameters includedDependencies and excludedDependencies,
similar to the existing included/excludedTasks parameters, one
can now choose which depenendencies to include into the build.

This is also useful for self-contained builds to include the built
depenendy into the bundle, instead of the unbuilt source read
from the file system (e.g. in case a transpile task is present).

## Pull Request Checklist


lib/builder/builder.js
```diff
@@ -207,6 +207,8 @@ module.exports = {
 	 * @param {string} parameters.destPath Target path
 	 * @param {boolean} [parameters.cleanDest=false] Decides whether project should clean the target path before build
 	 * @param {boolean} [parameters.buildDependencies=false] Decides whether project dependencies are built as well
+	 * @param {Array} [parameters.includedDependencies=[]] List of build dependencies to be included if buildDependencies is true
```
matz3: Array of which type? According to the code it's RegExp and strings, right?

kristian: yes, either RegExp, which will be matched against the project.metadata.name or otherwise compared with a strict comparison (===), so string yes!

kristian: any change required from my side? all the @param are only annotated with {Array}, or would you like to have it denoted elsewise? Thanks.

RandomByte: I think we should document this as `{string[]|RegExp[]}`

lib/builder/builder.js
```diff
@@ -251,21 +273,22 @@ module.exports = {
...
 			let depPromise;
 			let projectTasks = selectedTasks;
+
+			// Build dependencies in sequence as it is far easier to detect issues and reduces
+			// side effects or other issues such as too many open files
 			if (buildDependencies) {
-				// Build dependencies in sequence as it is far easier to detect issues and reduces
-				// side effects or other issues such as too many open files
-				depPromise = project.dependencies.reduce(function(p, depProject) {
+				depPromise = (project.dependencies || []).filter(projectFilter).reduce(function(p, depProject) {
```
RandomByte: `project.dependencies` must always be an array. A fallback should not be necessary.

See also https://github.com/SAP/ui5-project/pull/233

test/lib/builder/builder.js
```diff
@@ -462,37 +567,66 @@ test.serial("Cleanup", async (t) => {
 	t.deepEqual(executeCleanupTasksStub.callCount, 2, "Cleanup called twice");
 });
 
-const applicationATree = {
```
RandomByte: What happened here? Was applicationATree switched with libraryDTree?

lib/builder/builder.js
kristian: mhm, this doesn't really fit it... as it implies you can either have a string OR have a RegExp array... I adapted the guidance from https://jsdoc.app/tags-type.html, also see https://github.com/jsdoc/jsdoc/issues/714, and went for {Array.<string|RegExp>}

test/lib/builder/builder.js
kristian: - went with {Array.<string|RegExp>} for the JSDoc
- removed fallback from project.dependencies
- I wanted to use libraryDTree in applicationATree, so I moved libraryDTree above applicationATree and changed the dependency in applicaitonATree to include libraryDTree

RandomByte: 👍 nice

PR #386, opened 2019-12-27
[FIX] Avoid redundant bundle creation

codeworrior:
When a custom bundle has the same name as a built-in bundle (e.g.
foo/bar/library-preload.js), then the creation of the built-in bundle
can be skipped.


matz3:
superseded by https://github.com/SAP/ui5-builder/pull/741

PR #390, opened 2019-12-29
[FEATURE] Add generateResourcesJson task

codeworrior:
Introduces new task "generateResourcesJson" which creates the file `resources.json` for libraries, applications and themes.
The task analyses the resources and collects meta information and stores them in a file called `resources.json`.
This meta information includes module name, size, dependencies, includes, ...


codeworrior:
Note: failing test requires #391.

lib/processors/resourceListCreator.js
```diff
@@ -0,0 +1,646 @@
...
+
+			const p = name.lastIndexOf("/");
+			const prefix = name.substring(0, p + 1);
+			const basename = name.substring(p + 1);
+			if ( basename.match("[^/]*\\.library|Component\\.js|manifest\\.json") && !this._components.has(prefix) ) {
+				this._components.set(prefix, new ResourcesList(prefix));
+			}
+			// a .theme file within a theme folder indicates a library/theme package
+			// Note: ignores .theme files in library folders
+			if ( name.match("(?:[^/]+/)*themes/[^/]+/(?:\\.theming|library.source.less)") && !this._themePackages.has(prefix) ) {
```
codeworrior: In the Maven tooling, the `library.source.less` was not used as a marker, but the ui5-tooling does not create the expected `.theming` files. @matz3 What is the right approach here?  Should we create the `.theming` files or chose a different approach?

lib/processors/resourceListCreator.js
```diff
@@ -0,0 +1,646 @@
...
+			log.verbose(`    ${resource} (orphan)`);
+			if ( ++n > 20 ) {
+				log.verbose("		... (%d more)", unassigned.size - n);
+				break;
+			}
+		}
+		if ( options.failOnOrphans ) {
+			throw new Error("not all resources could be assigned to components");
+		}
+	}
```
codeworrior: Is there a guideline where errors should be thrown, processors or tasks?

lib/tasks/generateResourcesJson.js
```diff
@@ -0,0 +1,49 @@
...
+
+// const log = require("@ui5/logger").getLogger("builder:tasks:generateResourcesJson");
+const resourceListCreator = require("../processors/resourceListCreator");
+
+const DEFAULT_EXCLUDES = [
+	"!**/.DS_Store",
+	"!**/.eslintrc",
+	"!**/.eslintignore",
+	"!**/.gitignore"
+];
```
codeworrior: Do we have a common place for such excludes?

lib/processors/resourceListCreator.js
matz3: We need to create the files in order to support the Theming Engine. This becomes relevant when our main delivery is built with the UI5 Tooling.

For now the `library.source.less` as an indicator is IMO fine. We also use it within the theme build as the files are required for the build (which was not the case in the legacy Maven build).

lib/processors/resourceListCreator.js
```diff
@@ -0,0 +1,646 @@
...
+		this.resource = resource;
+	}
+
+	buffer() {
+		return this.resource.getBuffer();
+	}
+
+	getProject() {
+		return this.resource._project;
+	}
```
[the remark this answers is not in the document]
tobiasso85: it is not called! Will remove it.

lib/processors/resourceListCreator.js
```diff
@@ -0,0 +1,646 @@
...
+
+const DEBUG_RESOURCES_PATTERN = /-dbg((?:\.view|\.fragment|\.controller|\.designtime|\.support)?\.js|\.css)$/;
+
+// TODO share with module bundler
+
+function extractName(path) {
+	return path.slice( "/resources/".length);
+}
+
+class LocatorResource extends Resource {
```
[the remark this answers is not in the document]
tobiasso85: `resourceListCreator` and `moduleBundler` use it.

How about a subfolder in `processors` -> `processors/lib` for helpers and shared code between processors.
This was done the same way for `jsdoc` -> `processors/jsdoc/lib`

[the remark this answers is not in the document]
tobiasso85:
```log
  tasks › bundlers › generateStandaloneAppBundle.integration › integration: build application.b standalone

  Rejected promise returned by test. Reason:

  TypeError {
    message: 'resource.getProject is not a function',
  }

  › module.exports (lib/lbt/utils/escapePropertiesFile.js:14:27)
  › BundleBuilder.writePreloadModule (lib/lbt/bundle/Builder.js:432:30)
  › BundleBuilder.writePreloadFunction (lib/lbt/bundle/Builder.js:288:17)
  › BundleBuilder._createBundle (lib/lbt/bundle/Builder.js:151:4)
  › Object.module.exports [as task] (lib/tasks/bundlers/generateStandaloneAppBundle.js:84:2)
  › Object.build (lib/builder/builder.js:368:4)

```

lib/processors/resourceListCreator.js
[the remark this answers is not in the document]
tobiasso85:
```log
  tasks › bundlers › generateStandaloneAppBundle.integration › integration: build application.b standalone

  Rejected promise returned by test. Reason:

  TypeError {
    message: 'resource.getProject is not a function',
  }

  › module.exports (lib/lbt/utils/escapePropertiesFile.js:14:27)
  › BundleBuilder.writePreloadModule (lib/lbt/bundle/Builder.js:432:30)
  › BundleBuilder.writePreloadFunction (lib/lbt/bundle/Builder.js:288:17)
  › BundleBuilder._createBundle (lib/lbt/bundle/Builder.js:151:4)
  › Object.module.exports [as task] (lib/tasks/bundlers/generateStandaloneAppBundle.js:84:2)
  › Object.build (lib/builder/builder.js:368:4)

```

lib/processors/resourceListCreator.js
tobiasso85: I think the lowest level is here perfect because we have all the information there and typically the error should bubble all the way up

lib/processors/resourceListCreator.js
```diff
@@ -0,0 +1,697 @@
...
+
+
+const getResourceSize = async (resource) => {
+	try {
+		const content = await resource.getString();
+		return Buffer.byteLength(content, "utf8");
+	} catch (e) {
+		// ignore error
+	}
+
```
codeworrior, line 619: Isn't this the most expensive way to determine the size of the resource?

getString() has to read the resource, convert the UTF-8 byte stream to a Unicode16 char sequence; Buffer.byteLength then has to do the opposite operation just to determine the original UTF-8 byte length.

Overall performance when building `sap.ui.core` also looks very bad.

lib/tasks/generateResourcesJson.js
```diff
@@ -0,0 +1,49 @@
...
+		writtenResources.forEach((res) => {
+			if ( resources.indexOf(res) < 0 ) {
+				resources.push(res);
+			}
+		});
+	} else {
+		resources = await workspace.byGlob(["/resources/**/*.*", ...DEFAULT_EXCLUDES]);
+	}
+	const dependencyResources = await dependencies.byGlob("/resources/**/*.{js,json,xml,html,properties,library}");
+
```
codeworrior, line 40: I have no idea, why I added the resources from dependencies in my initial commit.

Thinking about it now, the resources.json is a library-local thing (or component-local thing) and should not include the dependencies. They have their own resources.json.

Wanted to verify the impact by building `sap.ui.layout` with dependencies. But building it failed in `sap.ui.core`, because `sap.ui.core` has modules outside its namespace -> orphans.

lib/processors/resourceListCreator.js
```diff
@@ -113,6 +124,21 @@ class ResourceInfo {
 		if ( this.included != null && this.included.size > 0 ) {
 			this.merged = true;
 		}
+		if (orig.size > 0) {
```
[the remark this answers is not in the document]
tobiasso85: good point :) just found it in the original sources.

lib/processors/resourceListCreator.js
tobiasso85: hm already counted the buffer length but that's not correct, have to find a better way.
The statInfo would be the best, but this needs to be updated when the sources are compressed

codeworrior: why / how was the buffer length not correct?

lib/processors/resourceListCreator.js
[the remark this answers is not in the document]
codeworrior: yes, there it's `orig.size >= 0`, too.

lib/processors/resourceListCreator.js
codeworrior: Ok, `Buffer.prototype.length` is the allocation size, not the content length. Too bad. To me that means, `@ui5/fs.Resource` needs to know the size, if neither `statInfo` nor `Buffer` are suitable to determine it.

tobiasso85: Maybe for
@ui5/fs.Resource
we can use the [stats.size](https://nodejs.org/api/fs.html#fs_stats_size) property for size
but we must ensure that each content modification will trigger a size update

codeworrior: Exactly, maintaining the (byte) size IMO should become a responsibility of the Resource, to get a proper abstraction.

Downside: when new content is provided as a string, this unfortunately requires either encoding the string as UTF-8 or using Buffer.byteLength.  Maybe that update of the size then should be done lazily.

lib/lbt/LocatorResourcePool.js
```diff
@@ -0,0 +1,24 @@
...
+		super();
+	}
+
+	prepare(resources) {
+		resources = resources.filter( (res) => !res.getStatInfo().isDirectory() );
+		// console.log(resources.map($ => $.getPath()));
+		return Promise.all(
+			resources.map(
+				(resource) => this.addResource( new LocatorResource(this, resource) )
+			).filter( (followUp) => followUp )
```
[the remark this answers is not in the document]
tobiasso85: done

lib/processors/resourceListCreator.js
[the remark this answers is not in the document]
tobiasso85: done

lib/processors/resourceListCreator.js
tobiasso85: I agree the Resource should be responsible.
If a string is provided it is anyways converted to a buffer and the buffer.byteLength is just a property so there should be no performance impact.

lib/processors/resourceListCreator.js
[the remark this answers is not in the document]
tobiasso85: the `original` one gets called, but the `copied` one didn't

lib/tasks/generateResourcesJson.js
tobiasso85: done with d97bbe8

lib/processors/resourceListCreator.js
tobiasso85: can use statinfo size with
https://github.com/SAP/ui5-fs/pull/253

lib/processors/resourceListCreator.js
matz3: We only do so in the internal theme build (which produces the current delivery for SAPUI5 / OpenUI5 (except the npm packages).
So I think we need to stay with `library.source.less` for the time being.

lib/tasks/generateResourcesJson.js
```diff
@@ -0,0 +1,49 @@
...
+	if (workspace.byGlobSource) { // API only available on duplex collections
+		resources = await workspace.byGlobSource(["/resources/**/*.*", ...DEFAULT_EXCLUDES]);
+		// HACK add resources from internal writer of workspace
+		const writtenResources = await workspace._writer.byGlob(["/resources/**/*.*", ...DEFAULT_EXCLUDES]);
+		writtenResources.forEach((res) => {
+			if ( resources.indexOf(res) < 0 ) {
+				resources.push(res);
+			}
+		});
+	} else {
```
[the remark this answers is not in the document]
RandomByte: `workspace.byGlob` also only returns resources of the current project. Searching with the namespace should be possible by now (I know your comment is from last year), since with v2.0 all projects are required to have a namespace.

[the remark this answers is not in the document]
codeworrior: And what about `sap.ui.core`? Yet another, hard-coded exception?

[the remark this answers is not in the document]
RandomByte: Yes, now I get it. Indeed for `sap.ui.core` the namespace `sap/ui/core` is insufficient. A glob pattern like to current one should be fine though.

Still, why do we use `byGlobSource` instead of `byGlob`? I thought we are also interested in generated files.

lib/tasks/generateResourcesJson.js
RandomByte: Not yet. Although I would not expect the last three in the `webapp` or `src` directories. Do we know of such cases?

Would it be possible to rather have an include on the expected file extensions?

package.json
```diff
@@ -21,7 +21,7 @@
 		"test": "npm run lint && npm run jsdoc-generate && npm run coverage",
 		"test-azure": "npm run lint && npm run jsdoc-generate && npm run coverage-xunit",
 		"lint": "eslint ./",
-		"unit": "rimraf test/tmp && ava",
+		"unit": "rimraf test/tmp && ava -v",
```
RandomByte, line 24: That's what the `unit-verbose` npm script one line below is supposed for. Or was this changed for some temporary test?

lib/processors/resourceListCreator.js
RandomByte: That's fine, but the error message should try to explain the context better. The call stack might get lost through catching and wrapping this error while bubbling up.

Some good examples:
* https://github.com/SAP/ui5-builder/blob/master/lib/processors/uglifier.js#L27
* https://github.com/SAP/ui5-builder/blob/master/lib/processors/bundlers/flexChangesBundler.js#L66

lib/tasks/generateResourcesJson.js
RandomByte: Not yet. Although I would not expect the last three in the `webapp` or `src` directories. Do we know of such cases?

Would it be possible to rather have an include on the expected file extensions?

package.json
tobiasso85: was just temporary, this change somehow slided in, will revert this line

lib/tasks/generateResourcesJson.js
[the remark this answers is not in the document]
tobiasso85: byGlobSource retrieves the debug files as well, byGlob does not retrieve them

lib/processors/resourceListCreator.js
tobiasso85: done

lib/tasks/generateResourcesJson.js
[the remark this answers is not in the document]
RandomByte: > byGlobSource retrieves the debug files as well.

Are you sure? I think that should not be the case unless they already exist in the source directory.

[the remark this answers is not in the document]
tobiasso85: you're right, byGlob retrieves all the files.

adjusted logic for copy from source such that it does not depend on the combination of byGlobSource and adding missing ones from byGlob on top.
https://github.com/SAP/ui5-builder/pull/390/commits/f346f3e2e827e97fab0f32160a85bb84c8409251

lib/lbt/LocatorResourcePool.js
```diff
@@ -0,0 +1,24 @@
...
+		);
+		// .then( () => {
```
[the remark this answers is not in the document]
tobiasso85: done

RandomByte:
I think I won't be able to review the functional correctness of this PR in coming week. I simply don't know enough about what this task does and should do.

I'm surprised by the amount of new test fixtures. But for the same reason as above I can't tell whether they are needed or not.

But I tried something that might be interesting: I removed the new task execution from `ApplicationBuilder.js`, `LibraryBuilder.js` and `ThemeLibraryBuilder.js`. With that change, the code coverage for `ResourceCollector.js`, `ResourceInfo.js` and `resourceListCreator.js` went from almost 100% to below 70% each. This gives me the impression that some of the new functions are only tested through integration tests and do not have dedicated unit tests:

Screenshot 2020-08-07 at 18 49 56

For new modules we typically try to have at least one unit test per function. First, to efficiently test a variety of inputs. And secondly, to have failing tests which already contain the name of faulty function instead of a more generic integration test which first needs to be debugged on its own to identify the issue.

RandomByte:
Building the openui5-sample-app with this change currently fails with an error:

```
⚠️  Process Failed With Error

Error Message:
resources.json generation failed with error: There are 1 resources which could not be assigned to components.

Stack Trace:
Error: resources.json generation failed with error: There are 1 resources which could not be assigned to components.
    at module.exports (/Users/d054485/SAPDevelop/dev/meta-ui5-tooling/ui5-builder/lib/processors/resourceListCreator.js:183:9)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async Object.build (/Users/d054485/SAPDevelop/dev/meta-ui5-tooling/ui5-builder/lib/builder/builder.js:370:4)
    at async Object.handleBuild [as handler] (/Users/d054485/SAPDevelop/dev/meta-ui5-tooling/ui5-cli/lib/cli/commands/build.js:97:2)
```

The errors logged during the build are:
```
ERR! lbt:resources:ResourcePool TypeError: Cannot read property 'type' of undefined
ERR! lbt:resources:ResourcePool     at new SapUiDefineCall (/ui5-builder/lib/lbt/calls/SapUiDefine.js:34:16)
ERR! lbt:resources:ResourcePool     at XMLCompositeAnalyzer.analyze (/ui5-builder/lib/lbt/analyzer/XMLCompositeAnalyzer.js:19:23)
ERR! lbt:resources:ResourcePool     at determineDependencyInfo (/ui5-builder/lib/lbt/resources/ResourcePool.js:71:35)
ERR! lbt:resources:ResourcePool     at processTicksAndRejections (internal/process/task_queues.js:93:5)
ERR! lbt:resources:ResourcePool     at async LocatorResourcePool.getModuleInfo (/ui5-builder/lib/lbt/resources/ResourcePool.js:175:11)
ERR! lbt:resources:ResourcePool     at async Promise.all (index 9)
ERR! lbt:resources:ResourcePool     at async module.exports (/ui5-builder/lib/processors/resourceListCreator.js:117:2)
ERR! lbt:resources:ResourcePool     at async Object.build (/ui5-builder/lib/builder/builder.js:370:4)
ERR! lbt:resources:ResourcePool     at async Object.handleBuild [as handler] (/ui5-cli/lib/cli/commands/build.js:97:2)
ERR! lbt:resources:ResourcePool  failed to parse or analyze %s: sap/ui/demo/todo/test/unit/AllTests.js TypeError: Cannot read property 'type' of undefined
ERR! lbt:resources:ResourcePool     at new SapUiDefineCall (/ui5-builder/lib/lbt/calls/SapUiDefine.js:34:16)
ERR! lbt:resources:ResourcePool     at XMLCompositeAnalyzer.analyze (/ui5-builder/lib/lbt/analyzer/XMLCompositeAnalyzer.js:19:23)
ERR! lbt:resources:ResourcePool     at determineDependencyInfo (/ui5-builder/lib/lbt/resources/ResourcePool.js:71:35)
ERR! lbt:resources:ResourcePool     at processTicksAndRejections (internal/process/task_queues.js:93:5)
ERR! lbt:resources:ResourcePool     at async LocatorResourcePool.getModuleInfo (/ui5-builder/lib/lbt/resources/ResourcePool.js:175:11)
ERR! lbt:resources:ResourcePool     at async Promise.all (index 9)
ERR! lbt:resources:ResourcePool     at async module.exports (/ui5-builder/lib/processors/resourceListCreator.js:117:2)
ERR! lbt:resources:ResourcePool     at async Object.build (/ui5-builder/lib/builder/builder.js:370:4)
ERR! lbt:resources:ResourcePool     at async Object.handleBuild [as handler] (/ui5-cli/lib/cli/commands/build.js:97:2)
ERR! lbt:resources:ResourcePool TypeError: Cannot read property 'type' of undefined
ERR! lbt:resources:ResourcePool     at new SapUiDefineCall (/ui5-builder/lib/lbt/calls/SapUiDefine.js:34:16)
ERR! lbt:resources:ResourcePool     at XMLCompositeAnalyzer.analyze (/ui5-builder/lib/lbt/analyzer/XMLCompositeAnalyzer.js:19:23)
ERR! lbt:resources:ResourcePool     at determineDependencyInfo (/ui5-builder/lib/lbt/resources/ResourcePool.js:71:35)
ERR! lbt:resources:ResourcePool     at processTicksAndRejections (internal/process/task_queues.js:93:5)
ERR! lbt:resources:ResourcePool     at async LocatorResourcePool.getModuleInfo (/ui5-builder/lib/lbt/resources/ResourcePool.js:175:11)
ERR! lbt:resources:ResourcePool     at async Promise.all (index 23)
ERR! lbt:resources:ResourcePool     at async module.exports (/ui5-builder/lib/processors/resourceListCreator.js:117:2)
ERR! lbt:resources:ResourcePool     at async Object.build (/ui5-builder/lib/builder/builder.js:370:4)
ERR! lbt:resources:ResourcePool     at async Object.handleBuild [as handler] (/ui5-cli/lib/cli/commands/build.js:97:2)
```

RandomByte:
I also did some benchmarking using:

## OpenUI5 [sap.m](https://github.com/SAP/openui5/tree/14917b0428a914447207666630ac24fccb688660/src/sap.m) library

| ui5-builder ref | Command | Mean [s] | Min [s] | Max [s] |
|:---|:---|---:|---:|---:|
| master (e7d4431) | `ui5 build` | 24.354 ± 2.094 | 21.244 | 26.661 |
| create-resources-json (18fb726) | `ui5 build` | 27.820 ± 1.174 | 26.073 | 30.163 |



## [OpenUI5 Sample App](https://github.com/SAP/openui5-sample-app/tree/380bec94a5ecd3247980251808614e580785706b) (+ 6 dependencies)

*(skipping the `generateResourcesJson` task for applications due to the [error above](https://github.com/SAP/ui5-builder/pull/390#issuecomment-670922389))*

| ui5-builder ref | Command | Mean [s] | Min [s] | Max [s] |
|:---|:---|---:|---:|---:|
| master (e7d4431) | `ui5 build --all` | 65.268 ± 2.128 | 60.574 | 68.159 |
| create-resources-json (18fb726) | `ui5 build --all` | 74.221 ± 1.947 | 70.642 | 77.307 |

---

Benchmark tool: [hyperfine](https://github.com/sharkdp/hyperfine) with one warmup
Hardware: MacBook Pro 2017 - i7 2,9GHz/16GB

codeworrior:
The error seen during the build of the openui5-sample-app should be fixed with https://github.com/SAP/ui5-builder/pull/491.

RandomByte:
@codeworrior and I discussed the [benchmark results](https://github.com/SAP/ui5-builder/pull/390#issuecomment-670927842):

* The resources.json is not consumed at runtime. Only some known tools make use of the file.
    * This means many UI5 Tooling consumer do not need it 
* Both measurements show an increase in runtime of about 14%.
* This task might fail for some projects that are currently working fine
    * For example a library with multiple namespaces would need a way to configure those (currently not possible). Otherwise the build will likely fail


Both, @codeworrior and I therefore think it would be better to skip this task by default, similar to the `generateManifestBundle` task. For this it needs to be added to the list here: https://github.com/SAP/ui5-builder/blob/master/lib/builder/builder.js#L54

Btw. should we document the structure of the resources.json somewhere? Do we expect other tools to make use of it? I'm not sure.

test/lib/builder/builder.js
```diff
@@ -277,6 +277,9 @@ test("Build application.g", (t) => {
 	const expectedPath = path.join("test", "expected", "build", "application.g", "dest");
 
 	return builder.build({
+		includedTasks: [
```
RandomByte, line 280: Is there any integration test left that does not execute the generateResourceJson task now?

tobiasso85: distributed it more

test/lib/builder/builder.js
```diff
@@ -86,14 +94,14 @@ test("Build application.a", (t) => {
 	return builder.build({
 		tree: applicationATree,
 		destPath,
-		excludedTasks: ["generateComponentPreload", "generateStandaloneAppBundle", "generateVersionInfo"]
+		excludedTasks: ["generateComponentPreload", "generateStandaloneAppBundle", "generateVersionInfo", "generateResourcesJson"]
```
RandomByte, line 97: "generateResourcesJson" is not default anymore and therefore does not need to be excluded here

test/lib/builder/builder.js
```diff
@@ -496,6 +554,9 @@ test("Build theme.j even without an library", (t) => {
 	const destPath = "./test/tmp/build/theme.j/dest";
 	const expectedPath = "./test/expected/build/theme.j/dest";
 	return builder.build({
+		includedTasks: [
```
RandomByte, line 557: Please keep at least one default theme build

test/lib/builder/builder.js
tobiasso85: done

test/lib/builder/builder.js
tobiasso85: done

codeworrior:
@RandomByte Reg. your comment:
> Btw. should we document the structure of the resources.json somewhere? Do we expect other tools to make use of it? I'm not sure.

To be honest, I didn't remember this, but a search in our Wiki revealed [an existing documentation in the SDK](https://ui5.sap.com/#topic/adcbcf8b50924556ab3f321fcd9353ea).

The documented entries had been the "public" ones at the time of writing. But I guess we could enhance the documentation with a few more properties. A JSON schema also could make sense, but low priority as no one would use it currently (and the file is not intended to be maintained manually).

lib/tasks/generateResourcesJson.js
```diff
@@ -0,0 +1,94 @@
...
+				"**/designtime/library-preload.designtime.js",
+				"**/library-preload.support.js"
+			].join(",")
+		});
+	}
+	return creatorOptions;
+}
+
+/**
+ * Task for creating a resources.json file, describing all productive build resources.
```
RandomByte, line 34: Should we add a link to the relevant documentation: https://openui5.hana.ondemand.com/#topic/adcbcf8b50924556ab3f321fcd9353ea ?

test/fixtures/library.n/src/library/n/Button.js
```diff
@@ -0,0 +1,56 @@
...
+			}
+		},
+		designtime: "library/n/designtime/Button.designtime",
+		renderer: {
+			apiVersion: 2,
+			render: function(oRm, oButton) {
+				sap.ui.requireSync("./changeHandler/SplitButton");
+				oRm.openStart("button", oButton);
+				oRm.class("libNBtnBase");
+				oRm.openEnd();
```
RandomByte, line 35: Why is there so much code in some of these fixtures? How is this relevant to the bundling/resources.json generation?

It's seemingly copied from somewhere. But I couldn't find it in the OpenUI5 repository. What's the origin of this code here and in the other fixtures?

RandomByte:
I cherry-picked the commits from https://github.com/SAP/ui5-builder/pull/491 onto this branch and tried building the openui5-sample-app again using `ui5 build --include-task=generateResourcesJson --verbose`.

It fails with a different error now:

```sh
[...]
verb lbt:analyzer:ComponentAnalyzer adding library dependency  sap/m/library.js false
verb lbt:analyzer:ComponentAnalyzer adding library dependency  sap/f/library.js false
verb lbt:analyzer:ComponentAnalyzer adding library dependency  sap/ui/unified/library.js false
verb lbt:analyzer:ComponentAnalyzer derived model implementation dependency  sap/ui/model/resource/ResourceModel.js
verb lbt:analyzer:ComponentAnalyzer derived model implementation dependency  sap/ui/model/json/JSONModel.js
verb lbt:resources:ResourceCollector   configured external resources filters (resources outside the namespace): (none)
verb lbt:resources:ResourceFilterList filetypes: undefined
verb lbt:resources:ResourceFilterList   sap-ui-core-dbg.js --> include: "sap-ui-core-dbg.js"
verb builder:processors:resourceListCreator     writing 'sap/ui/demo/todo/resources.json'
ERR! builder:processors:resourceListCreator resources.json generation failed because of unassigned resources: sap-ui-version.json
ERR! builder:builder Build failed in 523 ms
info builder:builder Executing cleanup tasks...

⚠️  Process Failed With Error

Error Message:
resources.json generation failed with error: There are 1 resources which could not be assigned to components.

Stack Trace:
Error: resources.json generation failed with error: There are 1 resources which could not be assigned to components.
    at module.exports (/ui5-builder/lib/processors/resourceListCreator.js:198:9)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async Object.module.exports [as task] (/ui5-builder/lib/tasks/generateResourcesJson.js:88:24)
    at async Object.build (/ui5-builder/lib/builder/builder.js:382:4)
    at async Object.handleBuild [as handler] (/ui5-cli/lib/cli/commands/build.js:97:2)
```

This seems to be related to `sap-ui-version.json`.

I also don't really like that the ultimate error message is the same [as before](https://github.com/SAP/ui5-builder/pull/390#issuecomment-670922389), even though it's seemingly caused by something else(?)  
This will be the error message people will open issues with. Currently it doesn't seem to be very helpful to us.

test/lib/builder/builder.js
```diff
@@ -464,7 +472,30 @@ test("Build application.j", (t) => {
...
+	const destPath = "./test/tmp/build/application.j/dest-resources-json";
+	const expectedPath = path.join("test", "expected", "build", "application.j", "dest-resources-json");
+
+	return builder.build({
+		includedTasks: [
+			"generateResourcesJson"
+		],
+		tree: applicationJTree,
+		destPath,
+		excludedTasks: ["createDebugFiles", "generateStandaloneAppBundle", "generateVersionInfo"]
```
RandomByte, line 491: If we would not exclude `generateVersionInfo` here, this test would reproduce the issue I reported for the openui5-sample-app: https://github.com/SAP/ui5-builder/pull/390#issuecomment-672058203

lib/tasks/generateResourcesJson.js
```diff
@@ -0,0 +1,94 @@
...
+				"sap-ui-core*.js",
+				"**/Component-preload.js",
+				"**/library-preload.js",
+				"**/library-preload-dbg.js",
+				"**/library-preload.json",
+				"**/library-all.js",
+				"**/library-all-dbg.js",
+				"**/designtime/library-preload.designtime.js",
+				"**/library-preload.support.js"
+			].join(",")
```
codeworrior, line 27: I thought I had commented on this earlier, but can't find it. 

The representation of the filters as comma separated lists is a legacy from the Maven tooling. In a Maven Mojo, those lists have been configured as comma separated lists. But here, in a JavaScript task API, this doesn't make much sense. I would rather switch to pure arrays. 

But as long as this is an internal API (as it is currently), it doesn't matter and we could refactor this later.

lib/tasks/generateResourcesJson.js
tobiasso85: done

test/fixtures/library.n/src/library/n/Button.js
tobiasso85: removed unnecessary code lines, I couldn't find the origin either.
I think the main purpose was to have dynamic dependencies nested in the ast

test/lib/builder/builder.js
tobiasso85: this seems related to the resource path:
"other resources": `/resources/application/j/manifest.json`
"sap-ui-version.json": `/resources/sap-ui-version.json`
--> the prefix `/resources/application/j/` is expected by the grouping logic

lib/lbt/analyzer/JSModuleAnalyzer.js
```diff
@@ -553,14 +598,21 @@ class JSModuleAnalyzer {
...
-		function onSapUiRequireSync(node, conditional) {
+		function onSapUiRequireSync(node, conditional, name) {
 			const args = node.arguments;
 			const nArgs = args.length;
 			const i = 0;
 
 			if ( i < nArgs ) {
 				if ( isString(args[i]) ) {
-					const moduleName = ModuleName.fromRequireJSName( args[i].value );
+					// resolve dependencies absolute: e.g. "library/n/changeHandler/SplitButton.js"
```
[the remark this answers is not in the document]
tobiasso85: my bad, didn't know that
```js
sap.ui.requireSync("./Control")
ui5loader-dbg.js:468 Uncaught Error: relative name not supported ('./Control.js'
```

lib/lbt/analyzer/JSModuleAnalyzer.js
```diff
@@ -404,17 +442,24 @@ class JSModuleAnalyzer {
 				} else if ( isMethodCall(node, CALL_REQUIRE_SYNC) || isMethodCall(node, CALL_SAP_UI_REQUIRE_SYNC) ) {
 					// recognizes a call to sap.ui.requireSync
 					info.setFormat(ModuleFormat.UI5_DEFINE);
-					onSapUiRequireSync(node, conditional);
+
+					onSapUiRequireSync(node, conditional, info.name);
```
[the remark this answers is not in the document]
tobiasso85: done :)
-> above

lib/lbt/resources/ResourcesList.js
```diff
@@ -0,0 +1,153 @@
+const ResourceInfo = require("./ResourceInfo");
```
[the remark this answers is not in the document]
tobiasso85: done
(was also surprised by the plural "Resources" + "List")

lib/lbt/resources/ResourceCollector.js
```diff
@@ -0,0 +1,350 @@
...
+/**
+ * Collects all resources in a set of resource folders or from an archive
+ * and writes a 'resources.json' file for each component or library
+ * that can be found in the resources.
+ *
+ * @since 1.29.0
+ */
+class ResourceCollector {
+	/**
+	 * Collects resources
```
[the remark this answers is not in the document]
tobiasso85: done

lib/processors/resourceListCreator.js
```diff
@@ -0,0 +1,202 @@
...
+		supportResources: DEFAULT_SUPPORT_RESOURCES_FILTER.join(","),
+		debugBundles: DEBUG_BUNDLES.join(",")
+	}, options);
+
+	const pool = new LocatorResourcePool();
+	await pool.prepare( resources );
+
+	const collector = new ResourceCollector(pool);
+	const visitPromises = resources.map((resource) => {
+		return resource.getSize().then((size) => {
```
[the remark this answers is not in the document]
tobiasso85: good point

lib/tasks/generateResourcesJson.js
tobiasso85: the rebase -> force push changed the history o0 maybe it was lost/hard to find due to that

lib/processors/resourceListCreator.js
```diff
@@ -0,0 +1,202 @@
...
+
+/**
+ * Adds resources.json itself to the list
+ *
+ * @param {ResourcesList} list resources list
+ * @param {string} prefix
+ * @returns {string} new content with resources.json entry
+ */
+function addResourcesJsonToList(list, prefix) {
+	// having the file size entry part of the file is a bit like the chicken egg scenario
```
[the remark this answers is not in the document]
tobiasso85: done

test/lib/builder/builder.js
tobiasso85: sap-ui-version.json was excluded with edbf669
It is not relevant for the resources.json and should not be considered.

RandomByte: Will other files with a path like `/resources/myfile.xyz` still break the build?

lib/tasks/generateResourcesJson.js
tobiasso85: the file extensions can be anything required by the application, from image to web resource, therefore it is hard to define a whitelist.

test/expected/build/library.n/dest/resources/library/n/designtime/Button.designtime.js
```diff
@@ -3,4 +3,4 @@
  * (c) Copyright 2009-xxx SAP SE or an SAP affiliate company.
```
RandomByte, line 3: All these comments `UI development toolkit for HTML5 (OpenUI5)` should only be present if the file is copied or directly derived from the OpenUI5 repository.

For example in this case the code in this file has no relation to the code OpenUI5 repository anymore. Hence the comment should be removed. Please also review the other new and changed files in this PR for this!

tobiasso85: done

RandomByte: I was not aware that this comes from the configured copyright string and is part of the test. Changing "UI development toolkit for HTML5 (OpenUI5)" to "UI5 Tooling" would have been ok as well in this case. Now we have fewer tests for multiline copyright strings, oh well...

tobiasso85: tests for `library.e` still have the multiline comment

test/lib/builder/builder.js
tobiasso85: yes this currently breaks it.
But this is typically the result of a custom task, "normal" resources should reside in `/resources/application/j` folder

lib/tasks/bundlers/generateLibraryPreload.js
```diff
@@ -48,7 +48,7 @@ function getBundleDefinition(namespace) {
 
 						// include only thirdparty that is very likely to be used
 						"sap/ui/thirdparty/crossroads.js",
-						"sap/ui/thirdparty/caja-htmlsanitizer.js",
+						"sap/ui/thirdparty/caja-html-sanitizer.js",
```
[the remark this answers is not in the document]
tobiasso85: done with 7d41128

test/expected/build/application.j/dest-resources-json/changes/coding/MyExtension.js
```diff
@@ -0,0 +1 @@
+sap.ui.define([],function(){return{}});
```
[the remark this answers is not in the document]
tobiasso85: separated additional tests into separate PR:
#504

tobiasso85 (approved):
LGTM

matz3 (approved):

lib/tasks/generateResourcesJson.js
codeworrior: Update: Analysis of XMLViews and fragments requires the resources from dependencies. That was one reason t odd them.

PR #392, opened 2019-12-30
[BREAKING] LibraryFormatter: Ignore manifest.json of nested apps

RandomByte:
If a library contains a nested application project or similar, do not
assume that projects manifest.json to be relevant for the library itself
in case it does not feature a manifest.json.
Instead try to identify such situations by checking whether the
.library file and the manifest.json are on the same directory level.

BREAKING CHANGE:
If a library contains both, a manifest.json and .library file, they must
either be located in the same directory or the manifest.json is ignored.
In cases where the manifest.json is located on a higher level or
different directory on the same level than a .library file, an exception
is thrown.


RandomByte:
I think it would make sense to add this to the 2.0 release.

PR #411, opened 2020-02-04
[FIX] Ensure proper handling of multi-byte characters in streams

matz3:
The stream encoding needs to be set in order to properly handle
multi-byte utf8 characters that are split between different chunks.

See:
https://nodejs.org/api/stream.html#stream_readable_setencoding_encoding

### DIFF
--- a/lib/processors/stringReplacer.js
+++ b/lib/processors/stringReplacer.js
@@ -14,8 +14,9 @@ const replaceStream = require("replacestream");
  */
 module.exports = function({resources, options}) {
 	return Promise.all(resources.map((resource) => {
-		const stream = resource.getStream()
-			.pipe(replaceStream(options.pattern, options.replacement));
+		let stream = resource.getStream();
+		stream.setEncoding("utf8");
+		stream = stream.pipe(replaceStream(options.pattern, options.replacement));
 
 		resource.setStream(stream);
 		return resource;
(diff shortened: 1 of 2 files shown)


lib/processors/stringReplacer.js
```diff
@@ -14,8 +14,9 @@ const replaceStream = require("replacestream");
  */
 module.exports = function({resources, options}) {
 	return Promise.all(resources.map((resource) => {
-		const stream = resource.getStream()
-			.pipe(replaceStream(options.pattern, options.replacement));
+		const stream = resource.getStream();
+		stream.setEncoding("utf8");
+		stream.pipe(replaceStream(options.pattern, options.replacement));
 
 		resource.setStream(stream);
```
codeworrior: as far as I can see, stream.pipe returns another pipe (the destination of the pipe operation). In the original code, that return value was set as stream in the resource and I assume the same has to be done here.

matz3: Totally makes sense. Thanks!

matz3:
It seems that the fix doesn't work for Node 8...

RandomByte:
Let's fix this with version 2.0

matz3:
Or we skip the test in Node 8?

RandomByte:
Great idea, let's do that 👍

RandomByte:
Related: https://github.com/SAP/ui5-server/pull/280

PR #416, opened 2020-02-20
[FIX] ApplicationBuilder: Fix pattern to glob for .library files

svbender:
To generate the VersionInfo, the used glob pattern should only match .library files within the resource folder. Otherwise it is possible to find additional .library files, e.g. from unit tests.


matz3:
@svbender it seems that you need to sign the CLA, right?

PR #430, opened 2020-03-16
[BREAKING] Make namespace mandatory for application and library projects

svbender:
BREAKING CHANGE:
UI5 Project must be able to determine the project's namespace,
otherwise an error is thrown.

### DIFF
--- a/test/lib/types/library/LibraryFormatter.js
+++ b/test/lib/types/library/LibraryFormatter.js
@@ -282,13 +282,6 @@ test("getDotLibrary: reads correctly", async (t) => {
 	const expectedPath = path.join(myProject.path,
 		myProject.resources.configuration.paths.src, "library", "e", ".library");
 	t.deepEqual(fsPath, expectedPath, ".library fsPath is correct");
-
-	const {content: contentSecondCall, fsPath: fsPathSecondCall} = await libraryFormatter.getDotLibrary();
-	t.deepEqual(contentSecondCall.library.name, "library.e", ".library content has been read," +
-		"but should be cached now.");
-	const expectedPathSecondCall = path.join(myProject.path,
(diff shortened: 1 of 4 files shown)


test/lib/types/library/LibraryFormatter.js
```diff
@@ -284,6 +284,21 @@ test("getDotLibrary: reads correctly", async (t) => {
 	t.deepEqual(fsPath, expectedPath, ".library fsPath is correct");
 });
 
+test("getDotLibrary: reads correctly call again", async (t) => {
```
RandomByte: What is this testing exactly? The caching mechanism inside `getDotLibrary`?

svbender: Yes

RandomByte: But if I remove the caching here:
https://github.com/SAP/ui5-builder/blob/7773b4e8c6154330a16e6c7a5aff79ecc5579c63/lib/types/library/LibraryFormatter.js#L243-L245

The test will still be successful because it doesn't actually test whether something got cached (e.g. no second glob call).

---

Also, there is already tests for this. It just seem to be incorrect since its missing a second call: https://github.com/SAP/ui5-builder/blob/7773b4e8c6154330a16e6c7a5aff79ecc5579c63/test/lib/types/library/LibraryFormatter.js#L344

tobiasso85: exactly

RandomByte: > exactly

...

So we gonna fix the existing test instead?

svbender: 👍

PR #439, opened 2020-03-26
[INTERNAL] Adapt tests to new project._isRoot flag

RandomByte:
Flag was added with https://github.com/SAP/ui5-project/pull/287
Tests are not failing. This is just for consistency reasons and to
prevent future issues.


RandomByte:
Rebased due to conflicts

PR #453, opened 2020-04-29
[INTERNAL] Add verbose logging to component preload

RandomByte:


lib/types/application/ApplicationBuilder.js
```diff
@@ -4,6 +4,7 @@ const {getTask} = require("../../tasks/taskRepository");
 class ApplicationBuilder extends AbstractBuilder {
 	addStandardTasks({resourceCollections, project, log}) {
 		if (!project.metadata.namespace) {
+			// TODO 3.0: Throw here
```
matz3, line 7: We missed that for 2.0, right?

RandomByte: Yes, and now we can't change it without possibly breaking direct consumers 😕

PR #460, opened 2020-05-18
[FIX] Align JSDoc template and scripts with OpenUI5 1.79

codeworrior:


codeworrior:
Remember to bring the style change back to openui5!

flovogt:
> Remember to bring the style change back to openui5!

Already done. Thx for noting.

PR #466, opened 2020-06-03
[FIX] Bundler: Improve error log messages

matz3:


tobiasso85 (commented):
having stack traces in the log will flood it, would it make sense to have the stack traces logged with debug/verbose? and the error message with error?

tobiasso85 (approved):
LGTM

PR #477, opened 2020-07-06
[INTERNAL] lib/processors/jsdoc/lib/ui5 plugin.js: quote only empty s…

flovogt:
This PR removes the double quotes in the default value of the `@param` tag when a default values is of type string (e.g. enums) but has a non empty value.


flovogt:
Close this PR because another PR will cover the described issue.

Issue #404, opened 2020-01-24
Preload building of Fiori element-based apps produce verbose log content


petermuessig:
## Expected Behavior
Build is successful without error logging.

## Current Behavior
Build is logging a verbose error when analyzing the dependencies of apps using modules from Smart Templates / Fiori elements. 

## Steps to reproduce the issue
1.  Create a Fiori elements based app in SAP Web IDE
2. Run the build in SAP Web IDE or the ui5 build directly

## Context
- UI5 Module Version *(output of `ui5 --version` when using the CLI)*: `latest`
- Node.js Version: `independent`
- npm Version: `independent`
- OS/Platform: `independent`

## Affected components *(if known)*

- [X] [ui5-builder](https://github.com/SAP/ui5-builder)
- [ ] [ui5-server](https://github.com/SAP/ui5-server)
- [ ] [ui5-cli](https://github.com/SAP/ui5-cli)
- [ ] [ui5-fs](https://github.com/SAP/ui5-fs)
- [ ] [ui5-project](https://github.com/SAP/ui5-project)
- [ ] [ui5-logger](https://github.com/SAP/ui5-logger)

## Log Output / Stack Trace

```
Running task generateComponentPreload...
ERR! lbt:analyzer:SmartTemplateAnalyzer Error: resource not found in pool: 'sap/suite/ui/generic/template/ListReport/Component.js'
ERR! lbt:analyzer:SmartTemplateAnalyzer     at LocatorResourcePool.findResource (/tmp/webide.668.build/proj/node_modules/@ui5/cli/node_modules/@ui5/builder/lib/lbt/resources/ResourcePool.js:177:10)
```

codeworrior:
Looks like the app is missing a mandatory dependency.

petermuessig:
Yes, a dependency from the UI5 distribution layer which we need to publish. I just want to document this here to have a reference for customer tickets.

ShashankPincha:
Facing the same issue, unfortunately raised a ticket in @ui5/cli before I saw this ticket. https://github.com/SAP/ui5-cli/issues/294

Is there any solution?

petermuessig:
Hi @ShashankPincha , it is not a critical issue, just a logging that the SmartTemplate (Fiori elements) stuff is missing as a dependency and thus the builder logs this as missing modules. We are working right now to release the SAPUI5 dependencies on npmjs.com which will happen hopefully soon. The solution is in the pipeline... :-)

[...]

PR #479, opened 2020-07-15
[FIX] SmartTemplateAnalyzer: Do not throw in case missing dependency is expected

RandomByte:
In some cases it can be expected that SmartTemplateAnalyzer does not
have access to second-degree dependencies. In such cases, no error
should be thrown. Currently this causes misleading error logs, for
example in case of component preload generation for a Fiori Elements
project where no framework dependencies are available.

Resolves https://github.com/SAP/ui5-builder/issues/404

### DIFF
--- a/lib/lbt/analyzer/SmartTemplateAnalyzer.js
+++ b/lib/lbt/analyzer/SmartTemplateAnalyzer.js
@@ -119,7 +119,7 @@ class TemplateComponentAnalyzer {
 			}
 		} catch (err) {
 			if (this._pool.getIgnoreMissingModules() && err.message.startsWith("resource not found in pool")) {
-				log.verbose("Ignoring missing module as per ResourcePoool configuration: " + err.message);
+				log.verbose("Ignoring missing module as per ResourcePool configuration: " + err.message);
 			} else {
 				throw err;
 			}
--- a/lib/lbt/analyzer/SmartTemplateAnalyzer.js
+++ b/lib/lbt/analyzer/SmartTemplateAnalyzer.js
@@ -105,16 +105,24 @@ class TemplateComponentAnalyzer {
 
 	async _analyzeTemplateComponent(moduleName, pageConfig, appInfo) {
 		// console.log("analyzing template component %s", moduleName);
-		const resource = await this._pool.findResource(moduleName);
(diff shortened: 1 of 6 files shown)


lib/processors/bundlers/moduleBundler.js
```diff
@@ -112,16 +112,28 @@ class LocatorResourcePool extends ResourcePool {
...
  */
 module.exports = function({resources, options}) {
 //	console.log("preloadBundler bundleDefinition:");
 //	console.log(JSON.stringify(options.bundleDefinition, null, 4));
 
-	const pool = new LocatorResourcePool();
+	const bundleOptions = options.bundleOptions || {};
+
+	// Apply default bundleOptions
+	if (bundleOptions.optimize === undefined) {
```
matz3, line 125: I'm not completely sure, but I think the default handling is now different.

`optimize` will now always be set to true. Previously it has been set to `true` only in case no `bundleOptions` were provided at all.

RandomByte: Done in latest commit

lib/lbt/analyzer/SmartTemplateAnalyzer.js
```diff
@@ -105,16 +105,24 @@ class TemplateComponentAnalyzer {
...
+			const templateName = (pageConfig.component && pageConfig.component.settings &&
+									pageConfig.component.settings.templateName) || defaultTemplateName;
+			if ( templateName ) {
+				const templateModuleName = ModuleName.fromUI5LegacyName( templateName, ".view.xml" );
+				log.verbose("template app: add dependency to template view %s", templateModuleName);
+				appInfo.addDependency(templateModuleName);
+			}
+		} catch (err) {
+			if (this._pool.getIgnoreMissingModules() && err.message.startsWith("resource not found in pool")) {
+				log.verbose("Ignoring missing module as per ResourcePoool configuration: " + err.message);
```
matz3, line 122:
```suggestion
				log.verbose("Ignoring missing module as per ResourcePool configuration: " + err.message);
```

RandomByte: Done

PR #481, opened 2020-07-22
[FIX] generateLibraryPreload: Ignore missing modules

matz3:
In addition to #479 the generateLibraryPreload should also ignore missing modules when creating a bundle.
As only files from the workspace are bundled, it's not a problem when dependencies can't be found. Just for self-contained bundles, this is a problem.

See: https://github.com/SAP/ui5-builder/pull/479


test/lib/tasks/bundlers/generateLibraryPreload.js
```diff
@@ -0,0 +1,635 @@
...
+		ReaderCollectionPrioritized: t.context.ReaderCollectionPrioritizedStub
+	});
+
+	t.context.moduleBundlerStub = sinon.stub().resolves([]);
+	mock("../../../../lib/processors/bundlers/moduleBundler", t.context.moduleBundlerStub);
+
+	t.context.generateLibraryPreload = mock.reRequire("../../../../lib/tasks/bundlers/generateLibraryPreload");
+});
+
+test.afterEach(() => {
```
RandomByte, line 27:
```suggestion
test.afterEach.always(() => {
```

Issue #309, opened 2019-08-14
Build failed after `estraverse` upgraded to 4.3.0


zhangliangyi:
## Expected Behavior
The ui5 build works.

## Current Behavior
The ui5 build is failed.

## Steps to reproduce the issue
1. Execute ui5 build command and the error thrown

## Context
- UI5 Module Version *(output of `ui5 --version` when using the CLI)*: `{...}`
- Node.js Version: `8.12.0`
- npm Version: `5.6.0`
- OS/Platform: `SUSE Linux`
- Browser *(if relevant)*: `N/A`
- Other information: `N/A`

## Affected components *(if known)*

- [X] [ui5-builder](https://github.com/SAP/ui5-builder)
- [ ] [ui5-server](https://github.com/SAP/ui5-server)
- [ ] [ui5-cli](https://github.com/SAP/ui5-cli)
- [ ] [ui5-fs](https://github.com/SAP/ui5-fs)
- [ ] [ui5-project](https://github.com/SAP/ui5-project)
- [ ] [ui5-logger](https://github.com/SAP/ui5-logger)

## Log Output / Stack Trace
```
16:33:14 > Task :au-commonshell:au-commonshell-sap.sf.xlite.min-web:ui5Build FAILED
16:33:14 /tmpfs/jenkins/workspace/V4-PR_au-commonshell_PR-106-TZNXHPMO2RT2XDEYRYSBAOKX7MGM3KVWMEPWI2PLFL2S46VNLH2A/commonshell/au-commonshell-sap.sf.xlite.min-web/node_modules/@ui5/builder/lib/lbt/analyzer/JSModuleAnalyzer.js:179
16:33:14 			throw new Error(`unknown estree node type '${type}', new syntax?`);
16:33:14 			^
16:33:14 
16:33:14 Error: unknown estree node type 'ImportExpression', new syntax?
16:33:14     at Object.keys.forEach (/tmpfs/jenkins/workspace/V4-PR_au-commonshell_PR-106-TZNXHPMO2RT2XDEYRYSBAOKX7MGM3KVWMEPWI2PLFL2S46VNLH2A/commonshell/au-commonshell-sap.sf.xlite.min-web/node_modules/@ui5/builder/lib/lbt/analyzer/JSModuleAnalyzer.js:179:10)
```

Hi colleague,

I am from the SAP Sucessfactors, and we are using the ui5 builder in our daily build. Today we found an issue in ui5-builder and all the UI builds are failed.

[...]

PR #310, opened 2019-08-14
[FIX] Adapt to recent extension of estraverse's set of node types

codeworrior:
Fixes https://github.com/SAP/ui5-builder/issues/309

PR #489, opened 2020-08-06
[DEPENDENCY] Pin estraverse to v5.1.0

RandomByte:
Since the "self-protection" mechanism of our JSModuleAnalyzer (see https://github.com/SAP/ui5-builder/issues/309#issuecomment-521108883) is
highly dependent on the set of possible node types provided by
estraverse, we often saw problems in consuming projects after new
releases of estraverse. With estraverse@5.2.0 this is yet again the
case.

This change forces consumers to use a version of estraverse that is
supported by JSModuleAnalyzer (currently 5.1.0). Version updates will
only happen through pull requests created by dependabot.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
 		"escodegen": "^1.14.3",
 		"escope": "^3.6.0",
 		"esprima": "^4.0.1",
-		"estraverse": "^5.1.0",
+		"estraverse": "5.1.0",
 		"globby": "^11.0.1",
 		"graceful-fs": "^4.2.4",
 		"jsdoc": "^3.6.5",


RandomByte:
Estraverse added support for ChainExpression node (https://github.com/estools/estraverse/pull/113) which is not yet handled by ui5-builder leading to failing tests:

[`taskRepository: Failed to require task module for generateStandaloneAppBundle: unknown estree node type 'ChainExpression', new syntax?`](https://dev.azure.com/sap/opensource/_build/results?buildId=70035&view=logs&j=96b29ec3-7135-5fd2-32d9-9ff65489417f&t=d7fc28ce-7f14-5680-2b78-fc1a0ae9e409&l=1912)

RandomByte:
> If you want a quick solution

Yes



> I would rather stick to an older version of estraverse

Got it 👍 

It's amazing how I still can't wrap my head around this part of the JSModuleAnalyzer. My fix was solely based on the [last one](https://github.com/SAP/ui5-builder/pull/310) and hopes

RandomByte:
> Has it to be expected that the package-lock.json doesn't change? I thought, it is a super set of the information in the package.json, but maybe it isn't.

The lockfile intentionally does not define ranges for the direct dependencies of the project. Only fixed versions. In this case estraverse@5.1.0. 

The package-lock.json is a resolution of the package.json

codeworrior:
That was more or less clear to me. I just thought that the package-lock.json contains enough information to restore the package.json. But that seems not to be the case. 

Anyhow, I'm fine with the change.

PR #502, opened 2020-08-20
[INTERNAL] bundleBuilder: log verbose for json parse failure

tobiasso85:
When JSON content is parsed for minification, errors are logged using verbose level.
TheJSON content is not relevant, only for minification.

### DIFF
--- a/lib/lbt/bundle/Builder.js
+++ b/lib/lbt/bundle/Builder.js
@@ -411,8 +411,7 @@ class BundleBuilder {
 				try {
 					fileContent = JSON.stringify( JSON.parse( fileContent) );
 				} catch (e) {
-					log.verbose("Ignoring compression of JSON content. " +
-						"Failed to parse JSON file %s.", module);
+					log.verbose("Failed to parse JSON file %s. Ignoring error, skipping compression.", module);
 					log.verbose(e);
 				}
 			}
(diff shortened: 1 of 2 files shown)


lib/lbt/bundle/Builder.js
```diff
@@ -411,7 +411,8 @@ class BundleBuilder {
 				try {
 					fileContent = JSON.stringify( JSON.parse( fileContent) );
 				} catch (e) {
-					log.error(e);
+					log.verbose("error while parsing JSON file %s", module);
```
matz3, line 414: Maybe add a hint why the file is parsed and that the file is just bundled as is.

lib/lbt/bundle/Builder.js
```diff
@@ -411,7 +411,9 @@ class BundleBuilder {
 				try {
 					fileContent = JSON.stringify( JSON.parse( fileContent) );
 				} catch (e) {
-					log.error(e);
+					log.verbose("Ignoring compression of JSON content. " +
```
codeworrior, line 414: I would state this in the reverse order, first the reason, then the consequences:

 Failed to parse JSON .... Ignoring error, skipping compression.

tobiasso85: done with: a63ccd6

matz3 (approved):

PR #503, opened 2020-08-20
[FIX] generateLibraryPreload: Fix typo for caja html sanitizer

tobiasso85:
Fix path for caja-html-sanitizer:
This way it will be bundled and is part of the `library-preload.js`


matz3:
We should even call this [FIX].
It solves the problem that the caja-html-sanitizer.js is loaded individually instead of being part of the core library-preload.js

codeworrior:
We should fix this in our (SAP internal) voter middleware, too.

matz3 (approved):

PR #504, opened 2020-08-20
[INTERNAL] Add tests

tobiasso85:
Add tests to increase coverage
Fix naming in manifest.json file of library H

### DIFF
--- a/test/lib/types/themeLibrary/ThemeLibraryBuilder.js
+++ b/test/lib/types/themeLibrary/ThemeLibraryBuilder.js
@@ -42,5 +42,5 @@ test("tasks", async (t) => {
 		return themeLibraryBuilder.tasks[taskKey]();
 	});
 
-	t.is(asyncTasks.length, 4, "all 3 tasks should be added");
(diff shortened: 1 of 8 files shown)


test/lib/types/themeLibrary/ThemeLibraryBuilder.js
```diff
@@ -0,0 +1,46 @@
...
+		},
+		taskUtil: {
+			isRootProject: () => {
+				return true;
+			}
+		}
+	});
+	const asyncTasks = Object.keys(themeLibraryBuilder.tasks).map((taskKey) => {
+		return themeLibraryBuilder.tasks[taskKey]();
+	});
```
matz3, line 43: Whats the exact purpose of this? Just calling the function and expecting that no error is thrown?

matz3: Well, better than no test. I've changed the assertion and just called `build` which also invokes the task functions.

matz3 (approved):

PR #507, opened 2020-08-27
[INTERNAL] XMLTemplateAnalyzer: check expressions in coreRequire

tobiasso85:
If a `core:require` statement in a node contains an expression
e.g. `core:require="{= '{Handler: \'' + ${myActions > handlerModule} + '\'}'}"`
it is checked if the node resides within a `template` node.
If so the error will be logged with `verbose`.

### DIFF
--- a/lib/lbt/analyzer/XMLTemplateAnalyzer.js
+++ b/lib/lbt/analyzer/XMLTemplateAnalyzer.js
@@ -150,6 +150,7 @@ class XMLTemplateAnalyzer {
 
 		this.info = info;
 		this.conditional = false;
+		this.templateTag = false;
 		this.promises = [];
 		this.busy = true;
 
@@ -214,11 +215,13 @@ class XMLTemplateAnalyzer {
 		const localName = node.$ns.local;
 
 		const oldConditional = this.conditional;
+		const oldTemplateTag = this.templateTag;
 
 		if ( namespace === TEMPLATING_NAMESPACE ) {
 			if ( TEMPLATING_CONDITONAL_TAGS.test(localName) ) {
 				this.conditional = true;
 			}
+			this.templateTag = true;
 		} else if ( namespace === XHTML_NAMESPACE || namespace === SVG_NAMESPACE ) {
 
 			// ignore XHTML and SVG nodes
@@ -298,36 +301,17 @@ class XMLTemplateAnalyzer {
 
 		this._analyzeChildren(node);
 
-		// restore conditional state of the outer block
+		// restore conditional and templateTag state of the outer block
 		this.conditional = oldConditional;
+		this.templateTag = oldTemplateTag;
 	}
 
 	_analyzeChildren(node) {
 		if ( Array.isArray(node.$$) ) {
 			node.$$.forEach( (child) => {
-				// add parentNode reference to child to be able to walk up the parent chain
-				child.$parentNode = node;
 				return this._analyzeNode( child);
 			});
 		}
-	}
-
-	/**
-	 * Walks up the parent chain until fnMatch returns either <code>true</code> or there is no more parent
-	 *
-	 * @param {object} node xml node
-	 * @param {Function} fnMatch matcher function, gets node as argument and returns whether or not the current node matches
-	 * @returns {boolean} <code>true</code> if fnMatch finds a match, <code>false</code> otherwise
-	 * @private
-	 */
-	_findInParentChain(node, fnMatch) {
-		if (fnMatch(node)) {
-			return true;
-		}
-		if (node.$parentNode) {
-			return this._findInParentChain(node.$parentNode, fnMatch);
-		}
-		return false;
 	}
 
 	_analyzeCoreRequire(node) {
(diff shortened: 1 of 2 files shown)


lib/lbt/analyzer/XMLTemplateAnalyzer.js
```diff
@@ -304,14 +304,49 @@ class XMLTemplateAnalyzer {
...
 		}
+		return false;
 	}
 
 	_analyzeCoreRequire(node) {
 		const coreRequire = getAttributeNS(node, XMLVIEW_CORE_REQUIRE_ATTRIBUTE_NS);
 		let requireContext;
 		if ( coreRequire ) {
+			// expression evaluation
+			if (coreRequire.startsWith("{=")) {
```
matz3, line 338: I think we should not ignore those nodes in general. Otherwise we skip nodes that might be parsable or we still log an error in case there's a new syntax that doesn't start with `{=`.

lib/lbt/analyzer/XMLTemplateAnalyzer.js
```diff
@@ -304,14 +304,49 @@ class XMLTemplateAnalyzer {
...
+				// check if parent chain contains a "template" node
+				const fnMatch = (node) => {
+					return node.$ns.uri === TEMPLATING_NAMESPACE;
+				};
+				if (this._findInParentChain(node, fnMatch)) {
```
matz3, line 343: Another solution would be to re-use the `this.conditional` flag, which is already set within `_analyzeNode` when a template tag or its children are processed.

Maybe renaming the flag to some more meaningful name (something that includes templating) would make sense.

matz3 (changes requested):

lib/lbt/analyzer/XMLTemplateAnalyzer.js
codeworrior: But `conditional` __is__ the meaning: the node is visited at runtime only _conditionally_. The same property is used throughout the `JSModuleAnalyzer`.

tobiasso85: I would introduce a new flag `templateTag`, but the idea is good to have a flag which is set just before the children processing and reset afterwards

lib/lbt/analyzer/XMLTemplateAnalyzer.js
tobiasso85: yes I see

I just thought that other parsing errors will then get swallowed just like:
missing comma:

```xml
core:require="{
	Foo:'sap/ui/Foo'
	Bar:'myApp/Bar'
}"
```

lib/lbt/analyzer/XMLTemplateAnalyzer.js
tobiasso85: tackled with: 3269bb0

matz3: @codeworrior conditional is only used for the added dependencies, right?
In this case we have a different use case but based on the same check.

lib/lbt/analyzer/XMLTemplateAnalyzer.js
```diff
@@ -214,11 +215,13 @@ class XMLTemplateAnalyzer {
 		const localName = node.$ns.local;
 
 		const oldConditional = this.conditional;
+		const oldTemplateTag = this.templateTag;
 
 		if ( namespace === TEMPLATING_NAMESPACE ) {
 			if ( TEMPLATING_CONDITONAL_TAGS.test(localName) ) {
```
matz3, line 221: Sorry, I missed this if-clause. Therefore we really need to have two separate flags I think.

lib/lbt/analyzer/XMLTemplateAnalyzer.js
tobiasso85: conditional is only true for `template:if` and `template:repeat` for others like `template:with` it is false

lib/lbt/analyzer/XMLTemplateAnalyzer.js
tobiasso85: np :)

lib/lbt/analyzer/XMLTemplateAnalyzer.js
```diff
@@ -298,20 +301,33 @@ class XMLTemplateAnalyzer {
...
 	_analyzeCoreRequire(node) {
 		const coreRequire = getAttributeNS(node, XMLVIEW_CORE_REQUIRE_ATTRIBUTE_NS);
 		let requireContext;
 		if ( coreRequire ) {
+			// expression binding syntax within coreRequire and a template parent node
+			// These expressions cannot be parsed using parseJS and if within a template tag
+			// represent an expression binding which needs to be evaluated before analysis
+			// e.g. "{= '{Handler: \'' + ${myActions > handlerModule} + '\'}'}"
+			if ((coreRequire.startsWith("{=") || coreRequire.startsWith("{:=")) && this.templateTag) {
+				log.verbose("Ignoring core:require: Attribute contains an expression but is within a 'template' Node ", node.$ns.local);
```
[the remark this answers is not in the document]
tobiasso85: good point! how about:
"Ignoring core:require: Attribute of Node Button contains an expression binding and is within a 'template' Node"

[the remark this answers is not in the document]
matz3: 👍🏻

matz3 (approved):

PR #508, opened 2020-08-27
[INTERNAL] generateLibraryPreload: set requiresTopLevelScope correctly

tobiasso85:
Ensures that `requiresTopLevelScope` option in `raw-module` configuration from `.library` is preferred and used
when creating the library preload.

`Builder.js`: Use correct variable for `ModuleInfo` (exposedGlobals) to
generate the correct globals assignment in `library-preload.js`

### DIFF
--- a/lib/lbt/bundle/Builder.js
+++ b/lib/lbt/bundle/Builder.js
@@ -381,6 +381,14 @@ class BundleBuilder {
 	beforeWritePreloadModule(module, info, resource) {
 	}
 
+	/**
+	 *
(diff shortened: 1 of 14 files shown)


matz3:
LGTM but I'd like @codeworrior to also have a look into this.

codeworrior:
Sigh, double work. I was working on the same today morning (I had promised to do so, if you remember).

I'll push another commit to this PR here. Luckily, overlap on test/* is, ehm, _small_ :-)

matz3 (approved):

PR #509, opened 2020-08-28
[INTERNAL] LocatorResourcePool: pass through constructor arguments

tobiasso85:
`ignoreMissingModules` parameter is passed along now when creating
a `new LocatorResourcePool`

works in combination with #479 and #481

### DIFF
--- a/lib/lbt/resources/LocatorResourcePool.js
+++ b/lib/lbt/resources/LocatorResourcePool.js
@@ -1,12 +1,7 @@
 const ResourcePool = require("./ResourcePool");
 const LocatorResource = require("./LocatorResource");
 
-
 class LocatorResourcePool extends ResourcePool {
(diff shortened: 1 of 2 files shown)


lib/lbt/resources/LocatorResourcePool.js
```diff
@@ -3,8 +3,8 @@ const LocatorResource = require("./LocatorResource");
 
 
 class LocatorResourcePool extends ResourcePool {
-	constructor() {
-		super();
+	constructor(parameters) {
```
matz3, line 6: AFAIK you can also just completely remove the constructor.

tobiasso85: kk (y) that's even better, I give it a try

matz3 (approved):

PR #510, opened 2020-09-01
[INTERNAL] package.json: Increase test timeout to 2 minutes (xunit only)

RandomByte:
Hopefully this will reduce the number of random failures on Windows on
Azure. Presumably they are caused by some sort of timeout within AVA.


matz3:
Still failing. Let's try to increase the 10s timeout to 1 minute

matz3 (dismissed):
I'd say let's try whether this works in the long term

RandomByte:
I re-triggered your commit two times and all executions where successful. But let's try whether the serial flag makes any immediate difference.

RandomByte:
Seems to be stable as well. I would squash and merge this, so it's basically just your commit @matz3. What do you think?

matz3:
👍🏻

matz3 (approved):

PR #529, opened 2020-10-11
[FEATURE] Create designtime and support bundles for libraries

codeworrior:
When a bundle would be empty (not contain any modules), it will be skipped.

### DIFF
--- a/lib/tasks/bundlers/generateLibraryPreload.js
+++ b/lib/tasks/bundlers/generateLibraryPreload.js
@@ -2,10 +2,24 @@ const log = require("@ui5/logger").getLogger("builder:tasks:bundlers:generateLib
 const moduleBundler = require("../../processors/bundlers/moduleBundler");
 const ReaderCollectionPrioritized = require("@ui5/fs").ReaderCollectionPrioritized;
 
+const DEFAULT_FILE_TYPES = [
+	".js",
+	".control.xml", // XMLComposite
+	".fragment.html",
+	".fragment.json",
+	".fragment.xml",
+	".view.html",
+	".view.json",
+	".view.xml",
+	".properties",
+	".json"
+];
+
 function getDefaultLibraryPreloadFilters(namespace) {
 	return [
 		`${namespace}/`,
 		`!${namespace}/.library`,
+		`!${namespace}/*-preload.js`, // exclude all bundles
 		`!${namespace}/designtime/`,
 		`!${namespace}/**/*.designtime.js`,
 		`!${namespace}/**/*.support.js`,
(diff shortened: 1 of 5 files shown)


lib/tasks/bundlers/generateLibraryPreload.js
```diff
@@ -2,10 +2,24 @@ const log = require("@ui5/logger").getLogger("builder:tasks:bundlers:generateLib
...
+	".js",
+	".control.xml", // XMLComposite
+	".fragment.html",
+	".fragment.json",
+	".fragment.xml",
+	".view.html",
+	".view.json",
+	".view.xml",
+	".properties",
+	".json"
```
codeworrior, line 15: The general ".json" file type is NOT present in our Maven tooling. I've kept it nevertheless for compatibility with existing projects that use the ui5-tooling already.

lib/tasks/bundlers/generateLibraryPreload.js
```diff
@@ -247,22 +343,11 @@ module.exports = function({workspace, dependencies, options: {projectName}}) {
...
-						}).then(([bundle]) => {
-							if (bundle) {
-								// console.log(`${libraryNamespace}/library-preload.js bundle created`);
-								return workspace.write(bundle);
-							}
-						});
+						return createLibraryBundles(libraryNamespace, resources)
+							.then((results) => {
+								const bundles = Array.prototype.concat.apply([], results);
+								return Promise.all(bundles.map((bundle) => bundle && workspace.write(bundle)));
```
codeworrior, line 349: Due to the new `skipIfEmpty`, some of the entries in the `bundles` array might be `undefined`.

RandomByte:
I like this

lib/tasks/bundlers/generateLibraryPreload.js
```diff
@@ -2,10 +2,24 @@ const log = require("@ui5/logger").getLogger("builder:tasks:bundlers:generateLib
...
+	".view.xml",
+	".properties",
+	".json"
+];
+
 function getDefaultLibraryPreloadFilters(namespace) {
 	return [
 		`${namespace}/`,
 		`!${namespace}/.library`,
+		`!${namespace}/*-preload.js`, // exclude all bundles
```
matz3, line 22: This does not exclude the `designtime/library-preload.designtime.js` or `library-preload.support.js`. Is this intended?

codeworrior: But the excludes in line 23 and 24 both will exclude it, don't they?

matz3: Oh yeah, I missed that. It's friday 😄

RandomByte (approved):

PR #534, opened 2020-10-21
[PERF] BundleWriter: Improve performance

matz3:
The ENDS_WITH_NEW_LINE RegExp is quite slow on large strings.
Therefore the "endsWithNewLine" state is computed after every modification
to the "buf" in order to improve performance.

### Measurements

Based on sap.m build in https://github.com/SAP/openui5/commit/0d95920e3ef1f5712d467298e2eee86ec51e2f63

| Command | Mean [s] | Min [s] | Max [s] |
|:---|---:|---:|---:|
| 2a5194346279279a6fb28c7332245e1cc5360d63 (master)<br>`ui5 build --exclude-task="*" --include-task=generateLibraryPreload` | 15.077 ± 0.355 | 14.476 | 15.499 |
| 713acb3677ba31ea9ad88f5a2f97c58f48df72d3 (BundleWriter-perf)<br>`ui5 build --exclude-task="*" --include-task=generateLibraryPreload` | 12.393 ± 0.334 | 11.883 | 12.913 |

### DIFF
--- a/lib/lbt/bundle/BundleWriter.js
+++ b/lib/lbt/bundle/BundleWriter.js
@@ -3,7 +3,7 @@
 
 const NL = "\n";
 const ENDS_WITH_NEW_LINE = /(\r\n|\r|\n)[ \t]*$/;
-const SPACES_OR_TABS_ONLY = /^[ \t]*$/;
+const SPACES_OR_TABS_ONLY = /^[ \t]+$/;
 
 /**
  * A filtering writer that can count written chars and provides some convenience
@@ -85,4 +85,3 @@ class BundleWriter {
 }
 
 module.exports = BundleWriter;
-
--- a/lib/lbt/bundle/BundleWriter.js
+++ b/lib/lbt/bundle/BundleWriter.js
@@ -28,8 +28,10 @@ class BundleWriter {
 		for ( let i = 0; i < str.length; i++ ) {
 			writeBuf += str[i];
 		}
-		this.buf += writeBuf;
-		this.endsWithNewLine = ENDS_WITH_NEW_LINE.test(writeBuf);
+		if ( writeBuf.length >= 1 ) {
+			this.buf += writeBuf;
+			this.endsWithNewLine = ENDS_WITH_NEW_LINE.test(writeBuf);
+		}
 	}
 
 	writeln(...str) {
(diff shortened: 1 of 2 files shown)


RandomByte:
I like this

lib/lbt/bundle/BundleWriter.js
```diff
@@ -20,11 +20,17 @@ class BundleWriter {
...
 
 	write(...str) {
+		let writeBuf = "";
 		for ( let i = 0; i < str.length; i++ ) {
-			this.buf += str[i];
+			writeBuf += str[i];
+		}
+		if ( writeBuf.length >= 1 ) {
+			this.buf += writeBuf;
+			this.endsWithNewLine = ENDS_WITH_NEW_LINE.test(writeBuf);
```
codeworrior, line 33: This can give wrong results when write is called with blanks or tabs only. In that case, the old endsWithNewLine flag must be taken into account as well.

Correct would be something like

```
  ENDS_WITH_NEW_LINE.test(writeBuf) || (this.endsWithNewLine && !writeBuf.trim())
```

Instead of the trim(), one could use another RegExp /^[ \t]*$/

codeworrior (dismissed):

lib/lbt/bundle/BundleWriter.js
matz3: Sounds right 👍🏻 
As there were no dedicated unit tests, I've added one to cover the whole class.

lib/lbt/bundle/BundleWriter.js
```diff
@@ -2,7 +2,8 @@
 
 
 const NL = "\n";
-const ENDS_WITH_NEW_LINE = /(^|\r\n|\r|\n)[ \t]*$/;
+const ENDS_WITH_NEW_LINE = /(\r\n|\r|\n)[ \t]*$/;
+const SPACES_OR_TABS_ONLY = /^[ \t]*$/;
```
RandomByte, line 6:
```suggestion
const SPACES_OR_TABS_ONLY = /^[ \t]+$/;
```

Otherwise this also matches empty strings which is not what the variable name says 🧐

RandomByte (approved):

PR #535, opened 2020-10-22
[FEATURE] Tag bundles and ignore them in uglify task

RandomByte:


lib/tasks/bundlers/generateBundle.js
```diff
@@ -29,8 +32,11 @@ module.exports = function({workspace, dependencies, options: {projectName, bundl
 			},
 			resources
 		}).then((bundles) => {
-			bundles.forEach((bundle) => {
-				workspace.write(bundle);
+			return bundles.map((bundle) => {
```
matz3, line 35: Isn't `Promise.all` missing here?

matz3 (changes requested):

lib/tasks/bundlers/generateBundle.js
RandomByte: Done

matz3 (approved):

PR #542, opened 2020-11-09
[INTERNAL] CI: Add depcheck

matz3:


RandomByte:
> Unused devDependencies
> * coveralls

Yes, we can indeed remove this dependency now.

Also, what about adding the depcheck to the existing Coveralls GH action (ideally renaming that to something more general)? I don't think it needs to be executed in different node versions/platforms.

matz3 (commented):
👍🏻

RandomByte (approved):

PR #543, opened 2020-11-10
[INTERNAL] Add CodeQL Action

RandomByte:
I used the action's [template](https://github.com/SAP/ui5-builder/new/master?code_scanning_workflow_template=codeql&filename=.github%2Fworkflows%2Fcodeql-analysis.yml) and removed some parts since we only use one language (no need for a matrix) and no build is necessary (therefore no need for `github/codeql-action/autobuild`)

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,32 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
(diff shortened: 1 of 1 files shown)


RandomByte:
Can't really be tested in a PR because:

> Missing analysis for base commit 9feddff
> 
> This usually happens when you enable code scanning for the first time, or if the build failed. Since there is no analysis in the base branch, we do not know if there are any new alerts in this commit. If you enabled analysis on push, you should check for alerts for this branch in the security tab.

matz3 (approved):

PR #546, opened 2020-11-19
[FIX] JSModuleAnalyzer: Properly handle jQuery.sap.registerPreloadedModules calls

matz3:
Fixes: https://github.com/SAP/ui5-builder/issues/544


RandomByte:
I like this +1

codeworrior (approved):
I'm also fine with the fix, but I have the impression that there's a more fundamental problem behind. In old (pre-evo) sources, jQuery.sap.global should be treated as a raw module and should not be analysed at all (as it __is__ the loader and contains the API implementations).

Anyhow, improving the robustness of the analyser code is in general a good idea..

PR #551, opened 2020-11-24
[FIX] generateResourcesJson: Make resources.json generation deterministic

RandomByte:
Stop analyzing debug-resources if they correspond to a non-debug resource.

Only analyze the non-debug resource and  copy any dependency information to the corresponding debug-resource.

While this might not be the best approach given that analyzing the source ("debug") version of a resource might result in finding more dependencies (i.e. those mangled during minification), this change ensures that we generate the same resources.json, independently from the order of debug and non-debug resources supplied to the resourceListCreator processor.

Resolves https://github.com/SAP/ui5-tooling/issues/274

### DIFF
--- a/lib/lbt/resources/ResourceCollector.js
+++ b/lib/lbt/resources/ResourceCollector.js
@@ -190,7 +190,7 @@ class ResourceCollector {
 
 			if ( /(?:\.js|\.view\.xml|\.control\.xml|\.fragment\.xml)$/.test(name) ) {
 				if ( (!info.isDebug || debugBundleFilter.matches(name)) ) {
-					// Only analyze non-debug files which are not special debug bundles (like sap-ui-core-dbg.js)
+					// Only analyze non-debug files and special debug bundles (like sap-ui-core-dbg.js)
 					promises.push(
 						this.enrichWithDependencyInfo(info)
 					);
@@ -242,8 +242,12 @@ class ResourceCollector {
 			const nonDebugName = ResourceInfoList.getNonDebugName(dbgInfo.name);
 			const nonDbgInfo = this._resources.get(nonDebugName);
 			const newDbgInfo = new ResourceInfo(dbgInfo.name);
+
+			// First copy info of analysis from non-dbg file (included, required, condRequired, ...)
 			newDbgInfo.copyFrom(null, nonDbgInfo);
+			// Then copy over info from dbg file to properly set name, isDebug, etc.
 			newDbgInfo.copyFrom(null, dbgInfo);
+
 			this._resources.set(dbgInfo.name, newDbgInfo);
 		}
 	}
@@ -270,7 +274,7 @@ class ResourceCollector {
 		return filtersByComponent;
 	}
 
-	groupResourcesByComponents(options) {
+	groupResourcesByComponents() {
 		const orphanFilters = this.createOrphanFilters();
 		for (const resource of this._resources.values()) {
 			let contained = false;
--- a/lib/lbt/resources/ResourceCollector.js
+++ b/lib/lbt/resources/ResourceCollector.js
@@ -150,16 +150,25 @@ class ResourceCollector {
 		});
 	}
 
-	async determineResourceDetails({pool, debugResources, mergedResources, designtimeResources, supportResources}) {
+	async determineResourceDetails({
+		debugResources, mergedResources, designtimeResources, supportResources, debugBundles
+	}) {
 		const baseNames = new Set();
 		const debugFilter = new ResourceFilterList(debugResources);
 		const mergeFilter = new ResourceFilterList(mergedResources);
 		const designtimeFilter = new ResourceFilterList(designtimeResources);
 		const supportFilter = new ResourceFilterList(supportResources);
+		const debugBundleFilter = new ResourceFilterList(debugBundles);
 
 		const promises = [];
+		const nonBundledDebugResources = [];
 
 		for (const [name, info] of this._resources.entries()) {
+			if ( debugFilter.matches(name) ) {
+				info.isDebug = true;
+				log.verbose(`  found potential debug resource '${name}'`);
+			}
+
 			// log.verbose(`  checking ${name}`);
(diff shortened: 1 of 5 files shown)


RandomByte:
Obviously still missing dedicated tests for the added code and parameters.

We could not measure any immediate performance improvement.

We also discussed alternative implementations. Including one even more similar to how Maven does it, where we generate the resources.json information during the bundling (where no debug files exist yet) and enrich it with the debug file information at a later point. Either while they are being created or in a separate task. But this would be a bigger effort than what this PR does.

lib/lbt/resources/ResourcePool.js
```diff
@@ -177,14 +181,30 @@ class ResourcePool {
...
 		if ( info == null ) {
 			// console.log("analyzing ", name);
+			let dbgResource;
+			if (options && options.preferDebugResources) {
+				// If requested, try to retrieve the corresponding debug-resource (xyz-dbg.js) and pass it to
+				// determineDependencyInfo as well. Its dependency analysis will perform better with the
+				// not-minified content of a resource.
+				const dbgName = ResourceInfoList.getDebugName(name);
+				if (dbgName && (!options.debugBundleFilter || !options.debugBundleFilter.matches(dbgName))) {
+					dbgResource = this._resourcesByName.get(dbgName);
```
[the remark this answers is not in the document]
RandomByte: Maybe that's the best approach then? Rewriting the ResourcePool to tightly integrate with ui5-fs. Also, adding a new Class in ui5-fs to store metadata like the AST (or plain dependency information?) per-resource.

Every task modifying a resource should update this metadata or at least invalidate it in some way.

Then there would still be a final resource.json-generation task, but instead of analyzing resources it would use the metadata already available for most of them.

lib/lbt/resources/ResourceCollector.js
```diff
@@ -179,9 +194,17 @@ class ResourceCollector {
 			}
 
 			if ( /(?:\.js|\.view\.xml|\.control\.xml|\.fragment\.xml)$/.test(name) ) {
-				promises.push(
-					this.enrichWithDependencyInfo(info)
-				);
+				if ( (!info.isDebug || debugBundleFilter.matches(name)) ) {
```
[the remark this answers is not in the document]
RandomByte: As discussed, the mentioned change in ResourcePool has been retracted from this PR. Just to have it documented somewhere I created https://github.com/SAP/ui5-builder/pull/553

tobiasso85 (dismissed):
LGTM

lib/lbt/resources/ResourceCollector.js
```diff
@@ -180,9 +189,14 @@ class ResourceCollector {
...
+					// Only analyze non-debug files which are not special debug bundles (like sap-ui-core-dbg.js)
```
matz3, line 193:
```suggestion
					// Only analyze non-debug files and special debug bundles (like sap-ui-core-dbg.js)
```
or am I wrong?

lib/lbt/resources/ResourceCollector.js
```diff
@@ -253,19 +272,17 @@ class ResourceCollector {
 
 	groupResourcesByComponents(options) {
```
matz3, line 273: `options` could be removed completely, right?

test/lib/lbt/resources/ResourceCollector.js
```diff
@@ -133,6 +127,72 @@ test.serial("determineResourceDetails: view.xml", async (t) => {
...
+		"enrichWithDependencyInfo is called with debug bundle");
+});
+
+test.serial("determineResourceDetails: Debug files and non-debug files", async (t) => {
+	const resourceCollector = new ResourceCollector();
+
+	const enrichWithDependencyInfoStub = sinon.stub(resourceCollector, "enrichWithDependencyInfo")
+		.callsFake((resourceInfo) => {
+			// Simulate enriching resource info with dependency info to test whether it gets shared
+			// with the dbg resource alter on
```
matz3, line 150:
```suggestion
			// with the dbg resource later on
```

lib/lbt/resources/ResourceCollector.js
```diff
@@ -226,7 +235,17 @@ class ResourceCollector {
...
-		return Promise.all(promises);
+		await Promise.all(promises);
+
+		for (let i = nonBundledDebugResources.length - 1; i >= 0; i--) {
+			const dbgInfo = nonBundledDebugResources[i];
+			const nonDebugName = ResourceInfoList.getNonDebugName(dbgInfo.name);
+			const nonDbgInfo = this._resources.get(nonDebugName);
+			const newDbgInfo = new ResourceInfo(dbgInfo.name);
+			newDbgInfo.copyFrom(null, nonDbgInfo);
+			newDbgInfo.copyFrom(null, dbgInfo);
```
matz3, line 246:
```suggestion
			newDbgInfo.copyFrom(null, nonDbgInfo); // First copy info of analysis from non-dbg file (included, required, condRequired, ...)
			newDbgInfo.copyFrom(null, dbgInfo); // Then copy over info from dbg file to properly set name, isDebug, ...
```

matz3 (changes requested):

lib/lbt/resources/ResourceCollector.js
RandomByte: Absolutely right 👍

lib/lbt/resources/ResourceCollector.js
RandomByte: Yes 🙌

test/lib/lbt/resources/ResourceCollector.js
RandomByte: Done

lib/lbt/resources/ResourceCollector.js
RandomByte: Done

matz3 (approved):

codeworrior (approved):
LGTM

PR #556, opened 2020-11-30
[INTERNAL] Add manifestCreator to processors facade

codeworrior:


RandomByte (approved):

matz3 (changes requested):
manifestCreator has no JSDoc, so we should not make it available to be imported.
We should also add JSDoc and with that maybe re-check the public API.

RandomByte (dismissed):
Discarding my approval

RandomByte:
Done with https://github.com/SAP/ui5-builder/pull/648

PR #559, opened 2020-12-09
[INTERNAL] flexChangesBundler: Add SAP as user when layer is VENDOR

thuyboehm:
[INTERNAL] When the layer in the change file content is VENDOR, set support.user to SAP to fill the author in manage view.


Lonwyr (approved):

Lonwyr (approved):

lib/processors/bundlers/flexChangesBundler.js
```diff
@@ -37,6 +37,9 @@ module.exports = function({resources, options: {pathPrefix, hasFlexBundleVersion
 		const variantManagementChanges = [];
 
 		changesContent.forEach(function(content) {
+			if (content.layer === "VENDOR") {
+				content.support.user = "SAP";
+			}
```
codeworrior, line 42: Can't it be that layer is "VENDOR" and the user is not SAP? 
In other words: shouldn't the user only be set when there's no value for user yet?

thuyboehm: No, the VENDOR layer is only use by SAP. When the user is set, it have to override with SAP.

codeworrior: Thank you for the clarification.

codeworrior (approved):

PR #562, opened 2020-12-15
[FIX] Windows: Correctly handle project paths containing non-ASCII characters

RandomByte:
The `slash` module does not convert "extended-length" paths or paths
which contain non-ASCII characters as documented in the modules README:
https://github.com/sindresorhus/slash/blob/7a9158487d8046c13f2de1db7e3734d33bdbb1bd/readme.md#slash-

Since in our case the resulting POSIX paths are only used in our virtual
fs "@ui5/fs", these limitations are not relevent to us.

Therefore we can simply replace any backslashes.

Related to https://github.com/SAP/ui5-fs/pull/294
Resolves https://github.com/SAP/ui5-tooling/issues/469


matz3 (dismissed):

RandomByte:
Still missing: Tests for non-ASCII tests for ApplicationFormatter

RandomByte:
Confirmed that the LibraryFormatter tests would fail on Windows without the fix in https://dev.azure.com/sap/opensource/_build/results?buildId=115046&view=results

matz3 (approved):

RandomByte:
Now finally ready 🙂

matz3 (approved):

PR #564, opened 2020-12-16
[FEATURE] manifestBundler: Add support for sap.app/i18n/enhanceWith

matz3:
- [x] Support for `bundleName` (+ tests)
- [x] Should `sap.ui5/models/i18n/settings/enhanceWith` also be checked? -> No

### DIFF
--- a/lib/processors/bundlers/manifestBundler.js
+++ b/lib/processors/bundlers/manifestBundler.js
@@ -69,6 +69,7 @@ module.exports = ({resources, options: {namespace, bundleName, propertiesExtensi
 		function addI18nInfo(i18nPath) {
 			if (i18nPath.startsWith("ui5:")) {
 				log.warn(`Using the ui5:// protocol for i18n bundles is currently not supported ('${i18nPath}' in ${manifest.path})`);
+				return;
 			}
 			descriptorI18nInfos.set(
 				path.join(path.dirname(manifest.path), path.dirname(i18nPath)),
@@ -82,6 +83,8 @@ module.exports = ({resources, options: {namespace, bundleName, propertiesExtensi
 		// i18n section in sap.app can be either a string or an object with bundleUrl
 		if (typeof appI18n === "object") {
 			i18nFullPath = appI18n.bundleUrl;
+		} else {
+			i18nFullPath = appI18n;
 		}
 		if (!i18nFullPath) {
 			i18nFullPath = "i18n/i18n.properties";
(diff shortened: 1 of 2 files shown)


matz3:
> In addition: Check for ui5:// URLs for i18n resources and log or throw (to be discussed) an error indicating those are not supported.

IMO throwing an error would be incompatible. The protocol can also be used so I'd say we only log an error (as done with my recent commit)

RandomByte:
Then let's log a warning 👍

matz3:
Windows tests are failing. I think we should really use `path.posix` when creating virtual paths.

lib/processors/bundlers/manifestBundler.js
```diff
@@ -57,25 +57,65 @@ class I18nResourceList {
...
-		return {
-			path: path.join(path.dirname(manifest.path), path.dirname(i18nFullPath)),
-			rootName: path.basename(i18nFullPath, propertiesExtension)
-		};
+		if (i18nFullPath) {
+			addI18nInfo(i18nFullPath);
+		}
+
+		if (typeof appI18n === "object") {
+			if (Array.isArray(appI18n.enhanceWith)) {
```
RandomByte, line 100: Merge these two if statements into one?

lib/processors/bundlers/manifestBundler.js
```diff
@@ -57,25 +57,65 @@ class I18nResourceList {
...
 		// i18n section in sap.app can be either a string or an object with bundleUrl
-		if (typeof i18nFullPath === "object") {
-			i18nFullPath = i18nFullPath.bundleUrl;
-		}
-		if (!i18nFullPath) {
+		if (typeof appI18n === "object") {
+			if (appI18n.bundleUrl) {
+				i18nFullPath = appI18n.bundleUrl;
+			} else if (appI18n.bundleName) {
+				i18nFullPath = bundleNameToUrl(appI18n.bundleName, content["sap.app"]["id"]);
```
RandomByte, line 88: >  `content["sap.app"]["id"]`

Or should we rather use the namespace that has been passed into the processor?

I'm not sure, since they *should* be equal. But in the end the `namespace` will be used in the ui5-fs so maybe it's *more* correct?

RandomByte (changes requested):

lib/processors/bundlers/manifestBundler.js
matz3: Afaics there could be multiple manifest.json files, so we need to use the one from the manifest.

RandomByte: Ah, yes right!

RandomByte (approved):

PR #570, opened 2021-01-20
[FIX] Theme Build: Only process themes within library namespace

matz3:
Themes are expected to be within a "themes" folder directly below the
library namespace.
Other "themes" folders (e.g. within sub-directories) should be ignored
and not compiled/processed.

### DIFF
--- a/lib/types/library/LibraryBuilder.js
+++ b/lib/types/library/LibraryBuilder.js
@@ -152,6 +152,10 @@ class LibraryBuilder extends AbstractBuilder {
 		}
 
 		this.addTask("buildThemes", async () => {
+			// Only compile themes directly below the lib namespace to be in sync with the theme support at runtime
+			// which only loads themes from that folder.
+			const inputPattern = `/resources/${project.metadata.namespace}/themes/*/library.source.less`;
+
 			return getTask("buildThemes").task({
 				workspace: resourceCollections.workspace,
 				dependencies: resourceCollections.dependencies,
@@ -159,7 +163,7 @@ class LibraryBuilder extends AbstractBuilder {
 					projectName: project.metadata.name,
 					librariesPattern: !taskUtil.isRootProject() ? "/resources/**/*.library" : undefined,
 					themesPattern: !taskUtil.isRootProject() ? "/resources/sap/ui/core/themes/*" : undefined,
-					inputPattern: "/resources/**/themes/*/library.source.less"
+					inputPattern
 				}
 			});
 		});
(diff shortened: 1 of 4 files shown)


lib/types/library/LibraryBuilder.js
```diff
@@ -152,14 +152,18 @@ class LibraryBuilder extends AbstractBuilder {
 		}
 
 		this.addTask("buildThemes", async () => {
+			// Only compile themes directly below the lib namespace to be in sync with the theme support at runtime
+			// which only loads themes from that folder.
+			const inputPattern = `/resources/${project.metadata.namespace}/themes/*/library.source.less`;
```
RandomByte, line 157: Apparently the namespace can be undefined here (see https://github.com/SAP/ui5-builder/pull/570/files#diff-60671e579ddf44d5c765da690b34a67cec4ed58ac1d0fb93bab7d1befb98578fR7). So a check and fallback is needed

RandomByte (changes requested):
Do we deem this to be compatible because the runtime never loads themes from other paths anyways? Can't an application still load CSS from other paths?

What's the issue we are fixing here?

codeworrior:
> Do we deem this to be compatible because the runtime never loads themes from other paths anyways? Can't an application still load CSS from other paths?
> 
> What's the issue we are fixing here?

Yes, the missing runtime support is what makes us consider this as compatible. Libraries don't load themes from a location other than their namespace.

And the issue that we fix here is the unwanted (and sometimes broken) automatic triggering of a theme build just by the naming of a folder. The only convention for libraries is that their root namespace can contain a themes folder whose child folders contain the individual themes. And those will be built. See https://github.com/SAP/openui5/blob/master/docs/controllibraries.md#file-structure .

lib/types/library/LibraryBuilder.js
matz3: For `theme-library` I decided to rather stay with the existing pattern as we don't really know the namespaces. One solution would be to filter using the information of the available libraries, but that would be a bit more complex and requires that knowledge also within the `generateThemeDesignerResources` task.
I think for `theme-libraries` we can expect that it only contains `themes` folders that should be compiled.

RandomByte: Ok, makes sense! However, we still to add a check for the namespace here, otherwise you might end up with `/resources/undefined/themes/*/library.source.less` here

matz3: Done

RandomByte (approved):

PR #555, opened 2020-11-30
[FIX] manifestCreator: Only list components with corresponding 'embeddedBy'

matz3:

PR #575, opened 2021-01-27
[FIX] manifestCreator: 'embeds' should list all components

tobiasso85:
Field `embeds` in created manifest.json file now contains all nested
manifests, independent from the reverse field `embeddedBy` (`embeddedBy` 
points from an embedded manifest to the library's manifest).

Remove unreachable code (in `createSapUI5()` -> `getUI5Version()` function) .

Use `path.dirname` method to get directory name in `findEmbeddedComponents()` function.

Use string for version value in manifest (in `sectionVersion()` function).
e.g. "1.2.3" instead of the semver object's string representation.

It reverts changes which were introduced with #555
Successor of #554

### DIFF
--- a/lib/processors/manifestCreator.js
+++ b/lib/processors/manifestCreator.js
@@ -182,7 +182,7 @@ async function createManifest(libraryResource, libBundle, descriptorVersion, _in
 		// return undefined
 	}
 
-	async function createSapApp() {
+	function createSapApp() {
 		function hasManifest(componentPath, libraryPathPrefix) {
 			const manifestPath = componentPath + "/manifest.json";
 
@@ -640,7 +640,7 @@ async function createManifest(libraryResource, libBundle, descriptorVersion, _in
 
 	return {
 		"_version": descriptorVersion.toString(),
-		"sap.app": await createSapApp(),
+		"sap.app": createSapApp(),
 		"sap.ui": createSapUi(),
 		"sap.ui5": createSapUI5(),
 		"sap.fiori": createSapFiori(),
--- a/lib/processors/manifestCreator.js
+++ b/lib/processors/manifestCreator.js
@@ -183,7 +183,7 @@ async function createManifest(libraryResource, libBundle, descriptorVersion, _in
 	}
 
 	async function createSapApp() {
-		async function isEmbeddedByLibrary(componentPath, libraryPathPrefix) {
+		async function hasManifest(componentPath, libraryPathPrefix) {
 			const manifestPath = componentPath + "/manifest.json";
 
 			const manifestResource = libBundle.findResource(manifestPath.substring(libraryPathPrefix.length));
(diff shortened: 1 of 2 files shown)


lib/processors/manifestCreator.js
```diff
@@ -429,11 +385,6 @@ async function createManifest(libraryResource, libBundle, descriptorVersion, _in
 
 	function createSapUI5() {
 		function getUI5Version() {
-			let ui5Version;
-			if ( ui5Version != null ) {
-				return ui5Version;
-			}
```
matz3, line 435: I wonder where this code came from. It doesn't make sense, but maybe it needs to be fixed rather than being removed.

test/lib/processors/manifestCreator.js
```diff
@@ -422,6 +422,38 @@ test.serial("default manifest creation with special characters small app descrip
...
+			dependencies: [{
+				metadata: {
+					name: "sap.ui.core"
+				}
+			}]
+		}
+	};
+	t.is(errorLogStub.callCount, 0);
+
+	const options = {descriptorVersion: new Version("1.1.0")};
```
matz3, line 445: I wonder why the API requires a semver version instance instead of a plain string.
AFAICS the whole processor is not marked as public API so we should be able to do a cleanup.

As this is not related to the actual PR of fixing the `embeds` behavior, I'd be in favor of creating a separate PR.
Or is there a relation from the version handling change (toString) to the embeds change? At least I could not find any hint. Maybe to increase code coverage?

lib/processors/manifestCreator.js
```diff
@@ -177,75 +177,31 @@ async function createManifest(libraryResource, libBundle, descriptorVersion, _in
...
-				return false;
-			}
+			return true;
 		}
 
-		async function findEmbeddedComponents() {
+		function findEmbeddedComponents() {
 			const result = [];
-			const prefix = libraryResource.getPath().slice(0, - ".library".length);
+			const prefix = posixPath.dirname(libraryResource.getPath()) + "/";
```
matz3, line 199: Much better 👍🏻

matz3 (changes requested):

lib/processors/manifestCreator.js
tobiasso85: The code taken from the Java implementation had a member variable ui5version which could be set from the oustide via parameter. This is not the case here anymore so it can be removed.

test/lib/processors/manifestCreator.js
tobiasso85: inside the manifestCreator the SemVer object is used for version comparison, as api parameter (descriptorVersion) a string definitely makes sense.

matz3: Also see https://github.com/SAP/ui5-builder/pull/556 where we already discussed making the manifestCreator public and that we would re-check the API before doing so.

matz3 (approved):
LGTM. I've changed the PR title, as this is rather a fix than a feature

PR #580, opened 2021-02-03
[FIX] moduleBundler: Apply defaultFileTypes

matz3:
Use default if not provided in bundle definition.

In the past neither a default, nor the bundle definition configuration
was used, since the parameter was mistakenly expected on the bundle
options.

Also align the defaults for library- and component-bundles with Maven.

New moduleBundler defaults:  
* `.js`
* `.control.xml`
* `.fragment.html`
* `.fragment.json`
* `.fragment.xml`
* `.view.html`
* `.view.json`
* `.view.xm`

Component-bundles (Component-preload.js and sap-ui-custom.js) additionally include:
* `.properties`
* and via filters:
    * `${namespace}/**/manifest.json`
    * `${namespace}/changes/flexibility-bundle.json`
    * `${namespace}/changes/changes-bundle.json`

### DIFF
--- a/lib/tasks/bundlers/generateComponentPreload.js
+++ b/lib/tasks/bundlers/generateComponentPreload.js
@@ -68,6 +68,8 @@ module.exports = function({
 				const filters = [
 					`${namespace}/`,
 					`${namespace}/**/manifest.json`,
+					`${namespace}/changes/changes-bundle.json`,
+					`${namespace}/changes/flexibility-bundle.json`,
 					`!${namespace}/test/`
 				];
 
--- a/lib/tasks/bundlers/generateComponentPreload.js
+++ b/lib/tasks/bundlers/generateComponentPreload.js
@@ -67,8 +67,8 @@ module.exports = function({
 			const bundleDefinitions = allNamespaces.map((namespace) => {
 				const filters = [
 					`${namespace}/`,
-					`!${namespace}/test/`,
-					`!${namespace}/*.html`
+					`${namespace}/**/manifest.json`,
+					`!${namespace}/test/`
 				];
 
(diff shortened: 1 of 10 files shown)


RandomByte:
Paging @codeworrior:

Should `changes/flexibility-bundle.json` be part of a `Component-preload.js`? It is right now, but won't be after this change since we remove `.json` from the default file types.

lib/tasks/bundlers/generateComponentPreload.js
```diff
@@ -85,12 +85,25 @@ module.exports = function({
 				allNamespaces.forEach((ns) => {
 					if (ns !== namespace && ns.startsWith(`${namespace}/`)) {
 						filters.push(`!${ns}/`);
+						// Explicitly exclude manifest.json files of subcomponents since the general exclude above this
+						// comment only applies to the configured default file types, which do not include ".json"
+						filters.push(`!${ns}/**/manifest.json`);
```
RandomByte, line 90: Not sure whether this is our intention. But I guess it's a good solution for the time bing.

The exclude-regex generated by ResourceFilterList always applies the default file types and therefore won't exclude `manifest.json`:
```js
// !library/h/components/subcomponent3/ --> exclude: 
/^library\/h\/components\/subcomponent3\/(?:[^/]+\/)*[^/]*(?:\.js|\.control\.xml|\.fragment\.html|\.fragment\.json|\.fragment\.xml|\.view\.html|\.view\.json|\.view\.xml|\.properties)$/
```

matz3 (commented):
LGTM

RandomByte (approved):
LGTM

PR #581, opened 2021-02-04
[FIX] generateResourcesJson: Include dependencies of XML resources

matz3:
The XML analysis requires dependency resources to be able to list them
as "required" in the resources.json.

### DIFF
--- a/lib/processors/resourceListCreator.js
+++ b/lib/processors/resourceListCreator.js
@@ -127,10 +127,11 @@ function makeResourcesJSON(list, prefix) {
  * @private
  * @param {object} parameters Parameters
  * @param {module:@ui5/fs.Resource[]} parameters.resources List of resources
+ * @param {module:@ui5/fs.Resource[]} [parameters.dependencyResources=[]] List of dependency resources
  * @param {object} [parameters.options] Options
  * @returns {Promise<module:@ui5/fs.Resource[]>} Promise resolving with the resources.json resources
  */
-module.exports = async function({resources, options}) {
+module.exports = async function({resources, dependencyResources = [], options}) {
 	options = Object.assign({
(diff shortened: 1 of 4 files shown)


matz3:
Tests are missing

lib/processors/resourceListCreator.js
```diff
@@ -138,11 +138,13 @@ module.exports = async function({resources, options}) {
 		mergedResources: DEFAULT_BUNDLE_RESOURCES_FILTER,
 		designtimeResources: DEFAULT_DESIGNTIME_RESOURCES_FILTER,
 		supportResources: DEFAULT_SUPPORT_RESOURCES_FILTER,
-		debugBundles: DEBUG_BUNDLES
+		debugBundles: DEBUG_BUNDLES,
+		dependencyResources: []
```
RandomByte, line 142: Why make it part of the options and not of `parameters` itself (next to `resources`)?

RandomByte (changes requested):
LGTM. Just one question regarding the function signature

lib/processors/resourceListCreator.js
matz3: My idea was to keep the signature of having `resources` and `options` for processors, but it seems we already use other parameters in existing processors.
So having dependencyResources next to resources is probably better.

matz3: Done

matz3:
CI is broken since https://github.com/SAP/ui5-builder/commit/61e2e3c43f52eaee7c51c1298c08a07ccc656058

matz3:
CI should be solved via https://github.com/SAP/ui5-builder/pull/587

RandomByte (approved):

PR #582, opened 2021-02-05
[INTERNAL][WIP] More jsdoc template updates

codeworrior:


tobiasso85 (approved):
compared the changes with the ones from openui5/lib/jsdoc folder and all of them are present here
compared these files with the one from openui5/lib/jsdoc folder and they are equal

PR #588, opened 2021-02-23
[FIX] resourceListCreator: Include dependencies of subModules

matz3:
JIRA: CPOUI5FOUNDATION-310

### DIFF
--- a/lib/lbt/resources/ResourceCollector.js
+++ b/lib/lbt/resources/ResourceCollector.js
@@ -117,14 +117,24 @@ class ResourceCollector {
 				resourceInfo.dynRequired = true;
 			}
 
-			this._applyModuleInfoDependenciesToResourceInfo(resourceInfo, moduleInfo);
+			if ( moduleInfo.dependencies.length > 0 ) {
+				resourceInfo.required = resourceInfo.required || new Set();
+				resourceInfo.condRequired = resourceInfo.condRequired || new Set();
+				moduleInfo.dependencies.forEach((dep) => {
+					if ( moduleInfo.isConditionalDependency(dep) ) {
+						resourceInfo.condRequired.add(dep);
+					} else if ( !moduleInfo.isImplicitDependency(dep) ) {
+						resourceInfo.required.add(dep);
+					}
+				});
+			}
 
 			if ( moduleInfo.subModules.length > 0 ) {
 				resourceInfo.included = resourceInfo.included || new Set();
+				moduleInfo.subModules.forEach((mod) => {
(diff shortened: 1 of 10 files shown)


lib/lbt/resources/ResourceCollector.js
```diff
@@ -109,29 +109,40 @@ class ResourceCollector {
...
+					// Add subModule as included
+					resourceInfo.included.add(subModule);
+
+					// Try to inherit dependency info
+					let subModuleInfo;
+					try {
+						subModuleInfo = await this._pool.getModuleInfo(subModule);
+					} catch (err) {
+						// TODO: Decide on message and log level
+						log.warn("Missing module? " + subModule);
```
matz3, line 134: verbose

matz3: Done

lib/lbt/resources/ResourceCollector.js
```diff
@@ -109,29 +109,39 @@ class ResourceCollector {
...
+					// Try to inherit dependency info
+					let subModuleInfo;
+					try {
+						subModuleInfo = await this._pool.getModuleInfo(subModule);
+					} catch (err) {
+						log.verbose(`	missing submodule ${subModule} included by ${moduleInfo.name}`);
+					}
+					if (subModuleInfo) {
+						// Inherit dependencies
+						this._applyModuleInfoDependenciesToResourceInfo(resourceInfo, subModuleInfo);
```
matz3, line 137: Dependencies should not be added if they are already listed as "included". I'm working on a fix..

matz3: Also, a dependency should not be listed as required **and** condRequired

matz3:
Ready for review

RandomByte:
LGTM. But @codeworrior should definitely take a look at the dependency inheritance logic

codeworrior (approved):
It's a bit unfortunate that we have to repeat the dependency logic from ModuleInfo here, but ResourceInfo was kept rather simple (no logic, just a JSON-like object).

I'm also still struggling with the fact that eager dependencies of submodules become eager dependencies of the bundle (what if a submodule is not used at runtime?). Not sure whether this was a concrete requirement of the HAT packager or whether this behavior of Maven was "by accident".

But as the current focus is to get rid of differences between Maven and ui5-tooling, Im fine with the change.

matz3:
Thanks for the feedback! I agree. We should right now only focus on having the same behavior. Changes / improvements can then still be made later on if required.

PR #591, opened 2021-03-04
[FIX] versionInfoGenerator: fix hasOwnPreload flag

tobiasso85:
`hasOwnPreload` is only `true` if the library has `embeds` field set in manifest
and embedded manifest does not have a backlink (`embeddedBy`) to the library.


matz3:
I find `hasOwnPreload switched boolean` a bit hard to understand as a bug fix message.

Basically this is now fixing how the behavior of the `hasOwnPreload` flag was intended to be (as follow-up of https://github.com/SAP/ui5-builder/pull/554).

matz3 (approved):

PR #592, opened 2021-03-04
[INTERNAL] builder: add test for builder JSDoc parameter

tobiasso85:
Increase test coverage

### DIFF
--- a/test/lib/builder/builder.js
+++ b/test/lib/builder/builder.js
@@ -735,6 +735,28 @@ test.serial("Build library.j with JSDoc build only", (t) => {
 	});
 });
 
+test.serial("Build library.j with jsdoc flag set", (t) => {
+	const destPath = path.join("test", "tmp", "build", "library.j", "dest");
+	const expectedPath = path.join("test", "expected", "build", "library.j", "dest");
+
+	return builder.build({
+		tree: libraryJTree,
+		destPath,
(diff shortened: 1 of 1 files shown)


RandomByte:
A verb in the commit message would be nice 🙂 . Something like "Add".

🍒  on top would be some capital letters like "JSDoc"

RandomByte (changes requested):
Please be aware that the builder.js tests must write into separate target directories each. Currently, your new test `Build library.j with jsdoc flag set` uses the same target directory as `Build library.j with JSDoc build only`.

So it overwrites the build result of the previous test but apparently does not delete any files.

If you change the target to an isolated directory you will find that no JSDoc resources are generated, because the `excludedTasks: ["*"]` parameter overrules the `jsdoc: true` flat and excludes all JSDoc tasks.

The `jsdoc` flag [already excludes all unrelated tasks](.com/SAP/ui5-builder/blob/f7e22ba866bf6a61b0d2932ef18aad53127641fd/lib/builder/builder.js#L89-L98). So there's no need to use the `excludedTasks` parameter. After all, the UI5 CLI [doesn't use it as well](https://github.com/SAP/ui5-cli/blob/aa73811689489252aff7840ba6744af4e3284909/lib/cli/commands/build.js#L108).

PR #573, opened 2021-01-25
[FEATURE] Add Library/Component preload exclude configuration

matz3:
JIRA: CPOUI5FOUNDATION-233

PR #593, opened 2021-03-09
[FIX] LibraryFormatter: Do not throw for missing .library in legacy OpenUI5 theme libraries

RandomByte:
Exception handling was already in place for legacy OpenUI5 theme
libraries that still use type 'library' instead of 'theme-library':
https://github.com/SAP/ui5-builder/pull/437

This allowed them to still be formatted even though they are missing a
.library file.

However, the newly added preload exclude configuration fallback for
framework libraries also requires a .library file and did not make the
mentioned exception for legacy OpenUI5 theme libraries:
https://github.com/SAP/ui5-builder/pull/573

This fix adds that exception and the tests that were missing from
https://github.com/SAP/ui5-builder/pull/437

### DIFF
--- a/test/lib/types/library/LibraryFormatter.js
+++ b/test/lib/types/library/LibraryFormatter.js
@@ -36,6 +36,31 @@ const libraryETree = {
 	}
 };
 
+const legacyThemeLibPath = path.join(__dirname, "..", "..", "..", "fixtures", "theme.library.e");
+const legacyThemeLibTree = {
+	id: "@openui5/themelib_sap_bluecrystal",
+	version: "1.0.0",
+	path: legacyThemeLibPath,
+	dependencies: [],
+	_level: 0,
+	_isRoot: true,
+	specVersion: "2.0",
+	type: "library",
+	metadata: {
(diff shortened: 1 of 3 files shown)


test/lib/types/library/LibraryFormatter.js
```diff
@@ -396,6 +421,15 @@ test.serial("format: namespace resolution fails", async (t) => {
 	mock.stop("@ui5/logger");
 });
 
+test("format: legacy OpenUI5 theme library", async (t) => {
+	const myProject = clone(legacyThemeLibTree);
+	const libraryFormatter = new LibraryFormatter({project: myProject});
+	sinon.stub(libraryFormatter, "validate").resolves();
+
+	await t.notThrowsAsync(libraryFormatter.format(), "Does not throw for missing .library");
+	t.deepEqual(myProject.metadata.copyright, legacyThemeLibTree.metadata.copyright, "Copyright was not altered");
```
matz3, line 430: Probably a leftover from another test, right?

matz3 (approved):

PR #602, opened 2021-05-10
[FIX] lbt/Builder: Preserve comments in bundles

RandomByte:
Replace AST-to-code procedure based on escodegen, which apparently removes comments from the bundled source, with some basic string manipulation.

JIRA: CPOUI5FOUNDATION-260

### DIFF
--- a/lib/lbt/bundle/Builder.js
+++ b/lib/lbt/bundle/Builder.js
@@ -5,7 +5,6 @@
 const terser = require("terser");
 const {pd} = require("pretty-data");
 const esprima = require("esprima");
-const escodegen = require("escodegen");
 const {Syntax} = esprima;
 // const MOZ_SourceMap = require("source-map");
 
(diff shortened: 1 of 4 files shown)


lib/lbt/bundle/Builder.js
```diff
@@ -471,123 +461,77 @@ class BundleBuilder {
...
 	} catch (e) {
 		log.error("error while parsing %s: %s", moduleName, e.message);
 		return;
 	}
 
 	if ( ast.type === Syntax.Program &&
-			ast.body.length === 1 &&
-			ast.body[0].type === Syntax.ExpressionStatement ) {
-		// rewrite define to require.predefine
-		if ( targetBundleFormat.supportsNativeDefine() && isMethodCall(ast.body[0].expression, CALL_DEFINE) ) {
```
matz3, line 525: Has the `targetBundleFormat.supportsNativeDefine()` check be removed on purpose?
If so and in case we don't expect them to be used somewhere else I'd suggest to also remove the functions from the bundle formats. Currently both formats just return `false`.

matz3 (changes requested):

lib/lbt/bundle/Builder.js
matz3: After reading the previous commit messages, I guess I can answer the first question myself with a "Yes" 😄

RandomByte: Done 👍

matz3 (approved):

PR #606, opened 2021-05-27
[FIX][WIP] lbt/resources: Ignore debug files but prefer their content

RandomByte:
WIP commit to test in CI


RandomByte:
Obsolete. We solved this in a different way on task-level.

PR #598, opened 2021-04-19
[INTERNAL] JSDoc: Add validation for metadata 'library' information

matz3:

PR #609, opened 2021-06-03
[FIX] JSDoc: add project name to JSDoc configuration and api.json

codeworrior:
The JSDoc processor did not propagate the library name (project name) to the JSDoc configuration. As a consequence, the UI5 template (publish.js) could not properly validate the metadata property 'library' for control and element classes. It reported many false positives ("ERROR: specified library 'foo.bar' for class 'foo.bar.Control' doesn't match containing library 'undefined'").

### DIFF
--- a/lib/processors/jsdoc/jsdocGenerator.js
+++ b/lib/processors/jsdoc/jsdocGenerator.js
@@ -106,6 +106,7 @@ async function generateJsdocConfig({targetPath, tmpPath, namespace, projectName,
 			"ui5": {
 				"variants": ${JSON.stringify(variants)},
 				"version": "${version}",
+				"uilib": "${projectName}",
 				"jsapiFile": "${jsapiFilePath}",
 				"apiJsonFolder": "${apiJsonFolderPath}",
 				"apiJsonFile": "${apiJsonFilePath}"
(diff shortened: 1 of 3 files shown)


RandomByte (approved):

codeworrior:
@RandomByte thank you for the quick review!

lib/processors/jsdoc/jsdocGenerator.js
```diff
@@ -106,6 +106,7 @@ async function generateJsdocConfig({targetPath, tmpPath, namespace, projectName,
 			"ui5": {
 				"variants": ${JSON.stringify(variants)},
 				"version": "${version}",
+				"uilib": "${projectName}",
```
matz3, line 109: I think we should not use the projectName (= ui5.yaml metadata name) but the namespace.
Usually they are equal but the namespace is guaranteed to be correct as it is also used for the folder structure. The project name is mainly used for logging purposes.

See my abandoned change here: https://github.com/SAP/ui5-builder/pull/598/files#diff-03d08225d447cd18da349dd9928b9093a83e66d9a6aa66cee4e0a0038779a9c8

matz3:
Thanks! I only cherry-picked the openui5 change and missed to add my changes from #598.

lib/processors/jsdoc/jsdocGenerator.js
RandomByte: Hm, what does "correct" mean here anyways? If the "uilib" property shall contain the libraries name, the projectName is always correct imho.

matz3: Especially for that "library" check IMO it's important that it fits to the namespace of the library. If it would be something else (e.g. mylib instead of com.example.mylib), the check would fail and report false findings.

RandomByte: I see. In that case we should change it to the namespace

PR #611, opened 2021-06-07
[FIX] lib/processors/debugFileCreator: Add -dbg suffix only to files

svbender:
The debugFileCreator added the '-dbg' suffix to the first occurrence
in the path string which could also be a folder name. This change
ensures that the '.js' file extension is at the end of the string.

### DIFF
--- a/lib/processors/debugFileCreator.js
+++ b/lib/processors/debugFileCreator.js
@@ -14,7 +14,7 @@ const util = require("util");
  */
 module.exports = function({resources, fs}) {
 	const options = {
-		pattern: /((\.view|\.fragment|\.controller|\.designtime|\.support)?\.js)$/,
+		pattern: /((?:\.view|\.fragment|\.controller|\.designtime|\.support)?\.js)$/,
 		replacement: "-dbg$1"
 	};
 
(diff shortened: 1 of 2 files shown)


RandomByte:
Nice 🥳 

I like it!

lib/processors/debugFileCreator.js
```diff
@@ -14,7 +14,7 @@ const util = require("util");
  */
 module.exports = function({resources, fs}) {
 	const options = {
-		pattern: /((\.view|\.fragment|\.controller|\.designtime|\.support)?\.js)/,
+		pattern: /((\.view|\.fragment|\.controller|\.designtime|\.support)?\.js)$/,
```
RandomByte, line 17:
```suggestion
		pattern: /((?:\.view|\.fragment|\.controller|\.designtime|\.support)?\.js)$/,
```

To align this fully with the Maven implementation (see `UI5NamingConventions.java`), we could also make the second group non-capturing. We don't reference it anyways.

RandomByte (changes requested):

matz3 (approved):

RandomByte (approved):

PR #612, opened 2021-06-07
[FIX] JSDoc: Use namespace to derive uilib name for jsdoc generator

matz3:
Tests have been adopted to check whether the right information is used.

Follow-up of https://github.com/SAP/ui5-builder/pull/609

### DIFF
--- a/test/lib/processors/jsdoc/jsdocGenerator.js
+++ b/test/lib/processors/jsdoc/jsdocGenerator.js
@@ -10,7 +10,7 @@ test("generateJsdocConfig", async (t) => {
 		targetPath: "/some/target/path",
 		tmpPath: "/some/tm\\p/path",
 		namespace: "some/namespace",
-		projectName: "some.namespace",
+		projectName: "some.projectName",
 		version: "1.0.0",
 		variants: ["apijson"]
 	});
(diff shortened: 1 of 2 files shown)


RandomByte (approved):

test/lib/processors/jsdoc/jsdocGenerator.js
```diff
@@ -26,7 +26,7 @@ test("generateJsdocConfig", async (t) => {
 		.replace(backslashRegex, "\\\\");
 	const destinationPath = path.join("/", "some", "tm\\p", "path")
 		.replace(backslashRegex, "\\\\");
-	const jsapiFilePath = path.join("/", "some", "target", "path", "libraries", "some.namespace.js")
+	const jsapiFilePath = path.join("/", "some", "target", "path", "libraries", "some.projectName.js")
```
codeworrior, line 29: was it only a consequence of the change in line 13 that the name of the file changed or is i a conscious decision that it should be named after the project?

In Maven, the name was derived from the library name (see the deletion in https://git.wdf.sap.corp:8080/#/c/sapui5/sapui5.runtime/+/4441367/2/src/framework/maven/poms/library-parent/pom.xml ). 

BTW: in that change we abandoned the generation of the jsapi file as Eclipse is no longer supported since then.

RandomByte: > was it only a consequence of the change in line 13 that the name of the file changed

Yes, in the test, the project name now differs from the namespace (which can happen for some projects). We decided to use the projectName for the jsapi file. IIRC we preferred it for that since it is guaranteed to be unique per build, while the namespace may not.

> in that change we abandoned the generation of the jsapi file as Eclipse is no longer supported since then.

Should we plan that too for UI5 Tooling 3.0?

codeworrior: This mismatch with the Maven build still puzzles me, but okay...

Yes, we should plan the removal of Eclipse code-completion artefacts for 3.0.

matz3: I don't mind to use the `namespace` for the jsapi file if this ensures an alignment with Maven.

matz3: I clarified with Merlin that we will not change the file name. We don't expect that it is used anyways and until now nobody has complained about it.

RandomByte: Plus, we plan to remove it anyways...

codeworrior: Okay with me.

PR #617, opened 2021-06-15
[INTERNAL] JSDoc: Add check for copyright notice

matz3:
Checks for a copyright notice at the beginning of each file.
The comment doesn't need to be the first comment but it must be placed
before any code starts.

This implements the same check as the internal validation rule
DOCUMENTATION__COPYRIGHT_NOTICE_MISSING.

Adds a jsdoc exclude configuration for sap.ui.model.odata.datajs which
is a legacy redirect file that only exists in the internal core overlay.
It is configured to be excluded from the internal validation, so it also
needs to be excluded from the jsdoc build so that no error is caused.

JIRA: CPOUI5FOUNDATION-329

Cherry picked from https://github.com/SAP/openui5/commit/bcb511074319aca894a026379bd2979627fb69c7

### DIFF
--- a/lib/processors/jsdoc/lib/ui5/plugin.js
+++ b/lib/processors/jsdoc/lib/ui5/plugin.js
@@ -2652,6 +2652,27 @@ exports.astNodeVisitor = {
 				processPotentialEnum(node.expression.right, comment);
 			}
 
+		} else if ( node.type === Syntax.File ) {
+			// Check for copyright notice
+
+			let programBodyStart;
+			if ( node.program.body.length >= 1 ) {
+				programBodyStart = node.program.body[0].start
+			} else {
+				// File has no code at all
+				programBodyStart = Infinity;
+			}
+
+			const hasCopyrightComment = ( node.comments || [] ).some(function(commentBlock) {
+				return (
+					// Copyright comments must be at the top of the file, not between some code or at the end
+					commentBlock.end <= programBodyStart &&
+					/copyright|\(c\)|released under|license|\u00a9/.test(commentBlock.value)
+				);
+			});
+			if ( !hasCopyrightComment ) {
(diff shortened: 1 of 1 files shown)


codeworrior (approved):
The change itself looks good to me. But please rewrite the commit message before submit as it mentions a change in the openui5 code (.library) which is not present in this commit here.

codeworrior (dismissed):
U-Turn 1: 
Mhmm, after thinking again about it, I withdraw my approval. 

Isn't this a BREAKING change? Users of the ui5 tooling so far did not get an error when their files did not contain a copyright. The SAP internal UI5 validation did not run for their projects...

Internally, we have more control about whether a JSDoc error breaks the build or not.

U-Turn 2:
I tried it out now locally and to my surprise, the build did not break on errors. neither in the buld-sdk variant nor when using the @ui5/cli directly. 
So we can continue with this change, but should have another look at the exit code of our JSDoc build.

matz3:
Thanks. I've updated the commit message.

Yeah it seems that exit codes 0 and 1 are both resolving the promise. I couldn't find a hint why this was done in https://github.com/SAP/ui5-builder/pull/42.

matz3:
I've added a bullet-point to the list of breaking changes for 3.0

codeworrior:
> ... I couldn't find a hint why this was done in #42.

I think with older versions of JSDoc, we observed the same exit code for success and error. I can't say whether this was still the case for JSDoc 3.5.5 (with which we started in #42), but I guess that was the reason why we ignored the exit code.

codeworrior (approved):

matz3:
> Anyhow, we now rely on the exit code in the SAP internal build and should rely on it in the tooling as well (breaking -> 3.0)

Addressed via https://github.com/SAP/ui5-builder/pull/845

PR #622, opened 2021-06-23
[FIX] LibraryFormatter: Fix handling of paths containing special characters

RandomByte:
FS paths containing characters which can be interpreted as regular
expression metacharacters may lead to issues during the namespaces
detection of libraries.

Alternatively we could switch to using
https://www.npmjs.com/package/escape-string-regexp which we already use
elsewhere. But I think this case here can be kept simple.

Fixes https://github.com/SAP/ui5-tooling/issues/526

### DIFF
--- a/lib/types/library/LibraryFormatter.js
+++ b/lib/types/library/LibraryFormatter.js
@@ -270,26 +270,29 @@ class LibraryFormatter extends AbstractUi5Formatter {
 	}
 
 	getNamespaceFromFsPath(fsPath) {
-		// Transform path to POSIX and remove any trailing slashes
-		const posixFsPath = fsPath.replace(/\\/g, "/").replace(/\/$/, "");
+		// Regex to ensure trailing slash
+		const rOptionalTrailingSlash = /\/?$/;
 
-		// Remove base path from fsPath
-		const posixBasePath = this.getSourceBasePath(true);
+		// Transform path to POSIX and ensure a trailing slash for correct comparison
+		const posixFsPath = fsPath.replace(/\\/g, "/").replace(rOptionalTrailingSlash, "/");
+		const posixBasePath = this.getSourceBasePath(true).replace(rOptionalTrailingSlash, "/");
 
-		// Can match /library/src as well as /library/src/some/namespace
-		const basePathPrefixRegExp = new RegExp(`^${posixBasePath}/?`);
+		if (posixBasePath === posixFsPath) {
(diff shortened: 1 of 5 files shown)


lib/types/library/LibraryFormatter.js
```diff
@@ -276,10 +276,7 @@ class LibraryFormatter extends AbstractUi5Formatter {
 		// Remove base path from fsPath
 		const posixBasePath = this.getSourceBasePath(true);
 
-		// Can match /library/src as well as /library/src/some/namespace
-		const basePathPrefixRegExp = new RegExp(`^${posixBasePath}/?`);
-
-		if (!basePathPrefixRegExp.test(posixFsPath)) {
+		if (!posixFsPath.startsWith(posixBasePath)) {
 			if (posixBasePath === posixFsPath + "/") {
```
codeworrior, line 280: this condition suggests to me that `posixBasePath` always ends with a slash, but the test mocks the method with a return value without a trailing slash?

lib/types/library/LibraryFormatter.js
```diff
@@ -289,7 +286,11 @@ class LibraryFormatter extends AbstractUi5Formatter {
 			throw new Error(`Given file system path ${posixFsPath} is not based on source base ` +
 				`path ${posixBasePath}.`);
 		}
-		const namespacePath = posixFsPath.replace(basePathPrefixRegExp, "");
+		let namespacePath = posixFsPath.replace(posixBasePath, "");
+		if (namespacePath.startsWith("/")) {
+			// Remove any leading slash
+			namespacePath = namespacePath.substr(1);
+		}
```
codeworrior, line 293: What about the else case, isn't it an error? Depends on the question whether `posixBasePath` ends with a slash (then it's not an error) or not.

lib/types/library/LibraryFormatter.js
RandomByte: As discussed we should ensure that `posixBasePath` ends with a slash to make this comparison stable

lib/types/library/LibraryFormatter.js
RandomByte: Since we don't enforce the presence (or absence) of a trailing slash in the project path mapping, we need to deal with both cases. I updated my initial commit to deal with optional trailing slashes everywhere and make the comparison somewhat easier to understand

matz3 (approved):

PR #630, opened 2021-07-20
[FIX] XMLTemplateAnalyzer: Include document name for XML parsing errors

RandomByte:
Also properly wait for the asynchronous analyze function to prevent
unhandled promise rejections and possible race conditions.

Change wording of an unrelated error message.

Resolves https://github.com/SAP/ui5-tooling/issues/539


codeworrior (approved):
LGTM

PR #645, opened 2021-09-27
[FEATURE] Introduce build task replaceBuildtime

larskissel:
Introduce the build task 'replaceBuildtime' used in
the LibraryBuilder. It replaces the placeholder
${buildtime} with a current timestamp in the
format "YYYYMMDD-HHmm".
Please note that the ${buildtime} placeholder is
only replaced in sap/ui/Global.js by default.

JIRA: CPOUI5FOUNDATION-327


RandomByte:
Looking good! But I think this should be a [FEATURE] commit. Also, please mention in the commit message that the placeholder is only replaced in Global.js by default.

Also, an integration test would be good to make sure that the "Global.js"-pattern actually works in a real library. I think we can add another test in `test/lib/builder/builder.js`.

RandomByte (changes requested):
See comment

RandomByte:
The integration test should also ensure that Global-dbg.js contains the replacement as well!

test/expected/build/sap.ui.core-buildtime/dest/resources/sap/ui/core/Core-dbg.js
```diff
@@ -0,0 +1,5 @@
+/*!
+ * Some fancy copyright
+ */
+// replacement of $\{buildtime\} is only made in sap/ui/Global.js
+console.log('${buildtime}');
```
RandomByte, line 5: According to our `.editorconfig`, there should always be a newline at the end of *.js files. Apparently your IDE doesn't seem to use this configuration. Please add newlines where missing, or my IDE will do it next time I save this file, creating unnecessary diffs 😉

test/fixtures/sap.ui.core-buildtime/main/src/sap/ui/Global.js
```diff
@@ -0,0 +1,5 @@
+/*!
+ * ${copyright}
+ */
+// replacement of $\{buildtime\} is only made in sap/ui/Global.js
```
RandomByte, line 4:
```suggestion
// replacement of $\{buildtime\} is only made in sap/ui/Global.js and Global-dbg.js
```

The comment really confused me in the expected `Global-dbg.js`

RandomByte (changes requested):

RandomByte (approved):

RandomByte (approved):

matz3 (approved):

PR #648, opened 2021-10-04
[INTERNAL] Export build processor manifestCreator

larskissel:
Make the build processor 'manifestCreator' accessible
for external usage.

JIRA: CPOUI5FOUNDATION-299

### DIFF
--- a/index.js
+++ b/index.js
@@ -50,6 +50,10 @@ module.exports = {
 		 */
 		libraryLessGenerator: "./lib/processors/libraryLessGenerator",
 		/**
+		 * @type {import('./lib/processors/manifestCreator')}
+		 */
+		manifestCreator: "./lib/processors/manifestCreator",
+		/**
 		 * @type {import('./lib/processors/resourceCopier')}
 		 */
 		resourceCopier: "./lib/processors/resourceCopier",


larskissel:
Closing as duplicate of https://github.com/SAP/ui5-builder/pull/556

lib/processors/manifestCreator.js
```diff
@@ -667,7 +672,8 @@ module.exports = function({libraryResource, resources, options}) {
 		return Promise.resolve(null); // a fulfillment of null indicates that no manifest has been created
 	}
 
-	return createManifest(libraryResource, libBundle, options.descriptorVersion, options.include3rdParty)
+	return createManifest(libraryResource, libBundle, options.descriptorVersion, options.include3rdParty,
+		options.minUI5VersionRequired)
```
RandomByte, line 676: I would prefer a switch to simply omit the minUI5Version instead of _trying_ to resolve it. Maybe `omitMinUI5Version`?

In the ui5-server use case, there is no benefit from having the minUI5Version available in _some_ cases. So I think it should never be available in order to keep things simpler.

larskissel: As discussed, let's call it omitMinVersions because not only the minUI5Version but also the minVersion of dependencies could lead to issues. The added versionInfoGenerator test proves that missing (empty string) minVersion/minUI5Version info doesn't affect the result of a sap-ui-version.json

test/lib/processors/versionInfoGenerator.js
```diff
@@ -221,6 +221,90 @@ test.serial("versionInfoGenerator library infos with dependencies", async (t) =>
...
+				"sap.ui5": {
+					"dependencies": {
+						"minUI5Version": "",
+						"libs": {}
+					}
+				}
+			});
+		}
+	};
+	const myDep = {name: "my.dep", version: "1.2.3", libraryManifest: myDepManifest};
```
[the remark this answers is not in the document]
larskissel: Done

test/lib/processors/versionInfoGenerator.js
```diff
@@ -221,6 +221,67 @@ test.serial("versionInfoGenerator library infos with dependencies", async (t) =>
...
+			}
+		]
+	};
+	assertVersionInfoContent(t, oExpected, result);
+	t.is(t.context.infoLogStub.callCount, 1);
+	t.is(t.context.infoLogStub.getCall(0).args[0],
+		"Cannot find dependency 'my.dep' " +
+		"defined in the manifest.json or .library file of project 'lib.a'. " +
+		"This might prevent some UI5 runtime performance optimizations from taking effect. " +
+		"Please double check your project's dependency configuration.");
```
RandomByte, line 281: I just tested this PR in conjunction with https://github.com/SAP/ui5-server/pull/420 and found that this info log does not appear if no dependencies are maintained at all, since in that case no libraryInfos are supplied and therefore no manifestHints are evaluated in the versionInfoGenerator.

Overall I found that this PR does not change whether or not users will see this info logging. So we should be good to merge it.

PR #653, opened 2021-10-15
[FEATURE] Support build minification excludes

larskissel:
Supports the build configuration "minification excludes"
introduced in specVersion 2.6. This allows to exclude
JavaScript resources from minification (tasks: uglify and
createDebugFiles).

JIRA: CPOUI5FOUNDATION-405

### DIFF
--- a/test/lib/builder/builder.js
+++ b/test/lib/builder/builder.js
@@ -1690,8 +1690,7 @@ const libraryJTree = {
 		"configuration": {
 			"paths": {
 				"src": "main/src"
-			},
-			"propertiesFileSourceEncoding": "ISO-8859-1"
+			}
 		},
 		"pathMappings": {
 			"/resources/": "main/src"


larskissel:
Integration tests will be uploaded with the next commit.

lib/types/application/ApplicationBuilder.js
```diff
@@ -138,12 +138,17 @@ class ApplicationBuilder extends AbstractBuilder {
 			});
 		}
 
+		const minificationPattern = ["/**/*.js"];
+		if (["2.6"].includes(project.specVersion)) {
+			const minification = (project.builder && project.builder.minification) || {};
+			this.enhancePatternWithExcludes(minificationPattern, minification.excludes);
```
matz3, line 144: Should the pattern be prefixed with `/resources/` here?
The default pattern for applications is `/**/*.js`, so I'd expect the excludes to be possible for all paths, not just within `/resources/`.
Or do I missing something here? For libraries it totally makese sense as we only process `/resources/`.

matz3 (changes requested):

test/lib/builder/builder.js
```diff
@@ -1614,14 +1690,50 @@ const libraryJTree = {
 		"configuration": {
 			"paths": {
 				"src": "main/src"
-			}
+			},
+			"propertiesFileSourceEncoding": "ISO-8859-1"
```
RandomByte, line 1694: Why this change?

larskissel: Looks like I made copy&paste to the wrong object. Done.

PR #662, opened 2021-11-12
[INTERNAL] Disable special core handling if project contains bundle config

matz3:


RandomByte (approved):
We should execute a smoke test for this

matz3:
Smoke Test Build #731 started

matz3:
I've also manually verified the change with the openui5-sample-app and the current sap.ui.core library (1.97.0-SNAPSHOT).

matz3:
Smoke Test did not show any issues related to this change.

PR #669, opened 2021-11-25
[FIX] generateResourcesJson: Analyze debug bundles

matz3:


matz3:
Some more test cases should be added

RandomByte (approved):

PR #671, opened 2021-11-29
[FIX] Enable buildThemes for libraries without .library

larskissel:
Fixes that libraries using a "library.js" file but no ".library"
file can also build theming files.

Fixes: https://github.com/SAP/ui5-tooling/issues/582

### DIFF
--- a/lib/tasks/buildThemes.js
+++ b/lib/tasks/buildThemes.js
@@ -54,8 +54,12 @@ module.exports = async function({
 		*/
 	let availableLibraries;
 	if (pAvailableLibraries) {
-		availableLibraries = (await pAvailableLibraries).map((resource) => {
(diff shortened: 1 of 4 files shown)


matz3:
LGTM 👍🏻 

@RandomByte any opinion whether we want to have this in v2.x or only in v3.x? IMO we should do it in both versions.

RandomByte:
Merging this into master and releasing as a v2 patch is also fine with me. We should then do a cherry-pick to `next` with an `[INTERNAL]`-prefix as outlined here: https://github.com/SAP/ui5-tooling/blob/master/docs/Guidelines.md#implementing-changes-in-multiple-code-lines

matz3 (approved):

larskissel:
cherry-pick to `next` branch is prepared:
https://github.com/SAP/ui5-builder/pull/676

PR #685, opened 2022-01-18
[BREAKING] moduleBundler: Always default to `optimize: true`

matz3:
bundleOptions are often omitted. So document them as optional and always
default 'optimize' to true, even if bundleOptions are specified.

This also fixes a bug that appeared if no 'bundleOptions' are defined in the ui5.yaml (e.g. sap.ui.webc.common).


RandomByte (approved):

matz3:
👍🏻

RandomByte (dismissed):

RandomByte (approved):

PR #686, opened 2022-01-20
minify: Omit .js.map from build results / generateResourcesJson: Don't list resources omitted from build result

matz3:


codeworrior (commented):
LGTM. What about tests?

matz3:
I've updated the tests and added some new ones for generateResourcesJson that are part of the commit.
For the minify task, the test has been adopted/enhance in the "Update tests" commit (see https://github.com/SAP/ui5-builder/pull/686/commits/9dee65a1c64d1e17dca41828491298a8399ca756#diff-337db9da362f6f8c9d5d04e610cb11c03318d80732d87ac9096ddab733d81f43)

lib/tasks/generateResourcesJson.js
```diff
@@ -97,12 +97,20 @@ function getCreatorOptions(projectName) {
...
  * @param {string} parameters.options.projectName Project name
  * @returns {Promise<undefined>} Promise resolving with <code>undefined</code> once data has been written
  */
-module.exports = async function({workspace, dependencies, options: {projectName}}) {
-	const resources = await workspace.byGlob(["/resources/**/*"].concat(DEFAULT_EXCLUDES));
+module.exports = async function({workspace, dependencies, taskUtil, options: {projectName}}) {
+	let resources = await workspace.byGlob(["/resources/**/*"].concat(DEFAULT_EXCLUDES));
+
+	if (taskUtil) {
+		// Filter out resources that will be omitted from the build results
```
RandomByte, line 109: What about `dependencies` a few lines below?

Maybe make the dependencies parameter mandatory as advised in the comment (breaking) and create a combo locator as done in other tasks as well.

We probably need to promote the OmitFromBuildResult tag to a global one then: https://github.com/SAP/ui5-builder/blob/next/lib/builder/BuildContext.js#L6

RandomByte (changes requested):

lib/tasks/generateResourcesJson.js
matz3: Resources of dependencies are only used for dependency analysis so that required modules are listed correctly.
So at least for listing of the resources, this is not relevant.

But to be consistent, we should probably also filter the resources of the dependencies.

matz3: I saw the 3.0 comment but just wanted to solve this issue first. But it seems to make sense to just do that right away.

matz3: Btw, we still need to pass the resources separately, so AFAICS a combo locator can't be used.

matz3: Done. I will add the changes to the [FIX] commit once the PR is reviewed.

RandomByte: Sounds good 👍

RandomByte (approved):

matz3:
Squashed the latest commit into the existing generateResourcesJson fix commit.
Ready for final review 👍🏻

RandomByte (approved):

PR #689, opened 2022-01-21
[FIX] Bundler: Ensure reproducibility for bundles with multiple parts

matz3:
Sort modules in preload sections alphabetically.
This caused different bundle contents based on the order of modules that
the resolver provides, which is not deterministic.

Remove terser from size estimation as modules are now already optimized
if the 'optimize' bundle option is enabled.
This caused incorrect estimations about the bundle size which resulted
into size differences of the individual parts.


matz3:
Tests need to be updated / enhanced

PR #692, opened 2022-01-24
[FEATURE] Support ES2021 language features

matz3:
By upgrading espree to the latest version, ES2021 language features
are now supported.

JIRA: CPOUI5FOUNDATION-363

### DIFF
--- a/lib/lbt/analyzer/JSModuleAnalyzer.js
+++ b/lib/lbt/analyzer/JSModuleAnalyzer.js
@@ -64,6 +64,7 @@ const EnrichedVisitorKeys = (function() {
 		BreakStatement: [],
 		CallExpression: [], // special handling
 		CatchClause: ["param", "body"],
+		ChainExpression: [],
 		ClassBody: [],
 		ClassDeclaration: [],
 		ClassExpression: [],
@@ -125,8 +126,10 @@ const EnrichedVisitorKeys = (function() {
 		 * All properties in an object pattern are executed.
 		 */
 		ObjectPattern: [], // properties
+		// PrivateIdentifier: [], // will come with ES2022
 		Program: [],
 		Property: [],
+		// PropertyDefinition: [], // will come with ES2022
 		/*
 		 * argument of the rest element is always executed under the same condition as the rest element itself
 		 */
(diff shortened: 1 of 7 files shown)


RandomByte (commented):
LGTM, @codeworrior to approve

codeworrior:
I have to make up my mind how to properly handle chained expressions. Our check for conditional expressions wants to detect whether a certain code is always executed (eager) or not (conditional).

The following top level code
```
someName.memberA[(function() { sap.ui.require(["a/b/c"]); return "someProperty";}())] 
```
would eagerly depend on a/b/c.

But
```
someName.memberA?.[(function() { sap.ui.require(["a/b/c"]); return "someProperty";}())] 
```
would not. 

This no longer can be declared statically. Within a ChainExpression.expression, the optional flag of the contained MemberExpressions has to be taken into account. If it is true, then the property part becomes conditional.

Well, but to enable the new syntax elements, the change is both required and helpful.

codeworrior (approved):

matz3:
Thanks for the insights.

I also thought about handling for the ChainExpression.
But as e.g. `someName.memberA && sap.ui.require(["a/b/c"]);` will not result into a conditional dependency, I thought that we anyways have some cases where we could improve, so I didn't have a deeper look into all this.

What I forgot to mention:
ChainExpression is not part of the `Syntax`, so I first could run a build without any issues until I used a ChainExpression in the code, which resulted into an error in the JSModuleAnalyzer visit function ("Unhandled AST node type"). So maybe we need another check in the opposite direction, at least within a unit test, to ensure that we will add new node types to our code.

codeworrior:
Reg. the gaps in _espree.Syntax_ see https://github.com/eslint/espree/pull/532 (fixes https://github.com/eslint/espree/issues/531)

I'm waiting for OSPO to approve the (C)CLA.

pubmikeb:
Does this merge mean that UI5 Tooling 2.14.5 now supports the ES2021 language features?

matz3:
@pubmikeb this PR got merged into the `next` branch which will result into the `v3.0.0` release. There are prereleases available, as described in https://github.com/SAP/ui5-tooling/issues/506

As the espree update requires a higher node version, we are not able to port this into the v2.x codeline.

PR #695, opened 2022-01-26
[FEATURE] Generate source maps for bundles

RandomByte:
Supersedes https://github.com/SAP/ui5-builder/pull/282

Based on https://github.com/SAP/ui5-tooling/pull/583

JIRA: CPOUI5FOUNDATION-434


lib/processors/bundlers/moduleBundler.js
```diff
@@ -146,13 +148,20 @@ module.exports = function({resources, options: {bundleDefinition, bundleOptions}
...
 					// console.log("creating bundle as '%s'", "/resources/" + name);
-					const resource = new EvoResource({
+					const res = {};
+					res.bundle = new EvoResource({
 						path: "/resources/" + name,
 						string: content
 					});
-					return resource;
+					if (sourceMap) {
+						res.sourceMap = new EvoResource({
```
matz3, line 159: What about just pushing all resources (bundle + sourcemap) to the existing array and then picking the files by checking for the resource name / extension?

Right now our tasks don't seem to need a connection between bundle and sourcemap so it could just loop over the array and set the tags according to the file extension. To me this sounds like a feasible solution.

This way we could also stick to the existing API that we intended to be used by all kind of `processors`.

And adopting the existing tests (which I already started..), would be way easier.

matz3: I discussed this with @codeworrior and as the minifier processor already uses such an API, we will also stick with it here.

matz3: However, the JSDoc for @returns still needs to be adopted

lib/lbt/bundle/Builder.js
```diff
@@ -200,23 +215,35 @@ class BundleBuilder {
 		}
 	}
 
+	writeWithSourceMap(content) {
+		const transientSourceName = `${path.posix.basename(this._bundleName)}?inline-${this._inlineContentCounter++}`;
```
matz3, line 219: Is this now the final name, or do we have some other suggestions?
What about `?bundled-code-${i}` ?

matz3: Webpack uses an own protocol `webpack://` for sources that are part of the map (sourcesContent).
Built-in code is references e.g. as `webpack://<project-name>/webpack/bootstrap`.

matz3: Also see https://webpack.js.org/configuration/output/#outputdevtoolmodulefilenametemplate

RandomByte: I guess if we use a different protocol for those sources, they won't appear next to the bundle in the source tree of the dev tools?

matz3: @codeworrior do you have any preference?

lib/lbt/bundle/Builder.js
```diff
@@ -527,29 +693,90 @@ function rewriteDefine(targetBundleFormat, code, moduleName) {
...
+					// need to be moved by the amount of inserted characters
+					mapping[0] = mapping[0] + mappingChange.columnDiff;
+				}
+			});
+		});
+
+		sourceMap.mappings = encodeMappings(mappings);
+
+		// No need for file information in source map since the bundled code does not exist in any file anyways
+		delete sourceMap.file;
```
matz3, line 767: Is there a reason why this is only removed in the "predefine" use case, not when written as sap.ui.require.preload?
As `createTransientSourceMap` also doesn't add a "file", I would prefer to also get rid of it for the preload use case.

RandomByte: As discussed, the `file` attribute is optional and we don't really know what it's actually used for in current development tools. In this case however, there is actually no file that this source map belongs to, since it is a modified version of the source map of a minified resource.

In the case of transient source maps, they only belong to sections of a bundle resource. It might be incorrect to set the `file` attribute in that case, although I'm not sure about that.

In cases where we bundle whole resources without modifying their content, I think we should also take over their source maps without modification. I.e. without removing the `file` attribute.

lib/lbt/bundle/Builder.js
codeworrior: I guess I like 7 and 8 most, but not so much that I would say: go for it!

The good thing is, we can change the names, if we want. They're local to the source map, aren't they? 

I also had some doubts regarding the own protocol. Now that Matthias mentioned it, I remember that we encountered it in real life already. See internal Snippet #93307 (it shows a blank page - open dev tools and check the sources tree).

Chrome shows the webpack:// sources perfectly fine. Maybe the URL does not matter when there's a sources entry in the source map.

RandomByte: I also liked 7 and 8 the most. I tried 8 but then the name got truncated in a bad way. `bundle-code` seems to work pretty well:
Screenshot 2022-02-23 at 10 23 28
Screenshot 2022-02-23 at 10 26 33

lib/processors/bundlers/moduleBundler.js
RandomByte: Done

lib/lbt/bundle/Builder.js
matz3: I just wanted to mention how webpack does it, but I'm not really in favor of it.

`bundle-code` sounds good 👍🏻

matz3 (approved):

lib/tasks/bundlers/generateComponentPreload.js
```diff
@@ -148,12 +148,19 @@ module.exports = function({
 				});
 			}));
 		})
-		.then((processedResources) => {
-			return Promise.all(processedResources.map((resource) => {
+		.then((results) => {
+			const bundles = Array.prototype.concat.apply([], results);
```
matz3, line 152: I wonder why this is needed in here

lib/tasks/bundlers/generateStandaloneAppBundle.js
```diff
@@ -101,6 +101,35 @@ module.exports = async function({workspace, dependencies, taskUtil, options: {pr
...
+		for (let i = unoptimizedResources.length - 1; i >= 0; i--) {
+			const resourcePath = unoptimizedResources[i].getPath();
+			if (taskUtil.getTag(resourcePath, taskUtil.STANDARD_TAGS.IsDebugVariant)) {
+				const nonDbgPath = ModuleName.getNonDebugName(resourcePath);
+				if (!nonDbgPath) {
+					throw new Error(`Failed to resolve non-debug name for ${resourcePath}`);
+				}
+				unoptimizedModuleNameMapping[resourcePath] = nonDbgPath.slice("/resources/".length);
+			}
+		}
```
matz3, line 130: We have this code in three places now, so it would be better to refactor it

lib/processors/bundlers/moduleBundler.js
```diff
@@ -90,6 +90,8 @@ const log = require("@ui5/logger").getLogger("builder:processors:bundlers:module
  * @public
  * @typedef {object} ModuleBundleOptions
  * @property {boolean} [optimize=true] Whether the module bundle gets minified
+ * @property {boolean} [sourceMap] Whether to generate a source map file for the bundle.
+ *   Defaults to true if <code>optimize</code> is set to true
```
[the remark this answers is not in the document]
matz3: The default will be updated via https://github.com/SAP/ui5-builder/pull/709

lib/processors/bundlers/moduleBundler.js
```diff
@@ -101,18 +103,30 @@ const log = require("@ui5/logger").getLogger("builder:processors:bundlers:module
...
  * @alias module:@ui5/builder.processors.moduleBundler
  * @param {object} parameters Parameters
  * @param {module:@ui5/fs.Resource[]} parameters.resources Resources
  * @param {object} parameters.options Options
+ * @param {object} [parameters.options.moduleNameMapping]
+ 				Optional mapping of resource paths to module name in order to overwrite the default determination
  * @param {ModuleBundleDefinition} parameters.options.bundleDefinition Module bundle definition
  * @param {ModuleBundleOptions} [parameters.options.bundleOptions] Module bundle options
- * @returns {Promise<module:@ui5/fs.Resource[]>} Promise resolving with module bundle resources
+ * @returns {Promise<module:@ui5/builder.processors.MinifierResult[]>} Promise resolving with module bundle resources
```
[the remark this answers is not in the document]
matz3: Will be solved via https://github.com/SAP/ui5-builder/pull/709

lib/processors/bundlers/moduleBundler.js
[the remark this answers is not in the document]
matz3: I've added a bullet-point for adding 'sourceMap' to the specVersion 3.0 bundleOptions to https://github.com/SAP/ui5-tooling/issues/506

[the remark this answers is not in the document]
RandomByte: As discussed, adding this to the bundleOptions makes sense

lib/processors/minifier.js
```diff
@@ -34,10 +34,12 @@ const debugFileRegex = /((?:\.view|\.fragment|\.controller|\.designtime|\.suppor
  * @alias module:@ui5/builder.processors.minifier
  * @param {object} parameters Parameters
  * @param {module:@ui5/fs.Resource[]} parameters.resources List of resources to be processed
+ * @param {boolean} [parameters.addSourceMappingUrl=true]
```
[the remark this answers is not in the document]
RandomByte: Yes, I forgot about that. This should be changed then 👍

lib/tasks/bundlers/generateComponentPreload.js
RandomByte: I wonder the same 🤷

lib/tasks/bundlers/generateStandaloneAppBundle.js
RandomByte: As discussed, maybe we could have all bundling tasks call the `generateBundle`-task  or create a "helper" module in `lib/tasks/bundlers/` for this purpose 👍

lib/processors/minifier.js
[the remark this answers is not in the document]
matz3: Fixed via https://github.com/SAP/ui5-builder/pull/709

PR #700, opened 2022-02-09
[INTERNAL] Minifier: Use workerpool for terser

matz3:


RandomByte:
Rebased and refactored for ESM in https://github.com/SAP/ui5-builder/pull/875

PR #701, opened 2022-02-09
[FIX] XMLTemplateAnalyzer: Analyze core:require of FragmentDefinition

matz3:
The FragmentDefinition node should be skipped, but it may still contain
a core:require definition.
This is in line with the current runtime which also reads the require
and loads the module.

### DIFF
--- a/test/lib/lbt/analyzer/XMLTemplateAnalyzer.js
+++ b/test/lib/lbt/analyzer/XMLTemplateAnalyzer.js
@@ -326,6 +326,26 @@ test("integration: Analysis of an xml fragment", async (t) => {
 		"Implicit dependency should be added since a fragment is analyzed");
 });
 
+test("integration: Analysis of an xml fragment with core:require", async (t) => {
+	const xml = `
+	<core:FragmentDefinition
+		xmlns:core="sap.ui.core"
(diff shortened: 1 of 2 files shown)


test/lib/lbt/analyzer/XMLTemplateAnalyzer.js
```diff
@@ -326,6 +326,26 @@ test("integration: Analysis of an xml fragment", async (t) => {
...
+
+	const analyzer = new XMLTemplateAnalyzer(fakeMockPool);
+	await analyzer.analyzeFragment(xml, moduleInfo);
+	t.deepEqual(moduleInfo.dependencies,
+		[
+			"sap/ui/core/Fragment.js",
+			"sap/m/MessageToast.js"
+		], "Dependencies should come from the XML template");
+	t.true(moduleInfo.isImplicitDependency("sap/ui/core/Fragment.js"),
+		"Implicit dependency should be added since an XML Fragment is analyzed");
```
matz3, line 346: Is this actually expected? We don't do that for XMLViews, and I thought we are also not doing it for fragments. I can at least not think of a reason why it needs to be handled differently.

RandomByte (approved):

PR #706, opened 2022-02-21
[INTERNAL][FIX] check of minUI5Version as version number

thuyboehm:
The minUI5Version check is needed for generate a flexibility-bundle or a change-bundle. The check is now separated for the major and minor version number.


lib/tasks/bundlers/generateFlexChangesBundle.js
```diff
@@ -69,8 +69,9 @@ module.exports = async function({workspace, taskUtil, options: {namespace}}) {
 		`${pathPrefix}/changes/*.{change,variant,ctrl_variant,ctrl_variant_change,ctrl_variant_management_change}`);
 	if (allResources.length > 0) {
 		const version = await readManifestMinUI5Version();
+		const versionArray = version.split(".");
 		let hasFlexBundleVersion = false;
-		if (parseFloat(version) >= 1.73) {
+		if (versionArray.length >= 2 && parseFloat(versionArray[0]) >= 1 && parseFloat(versionArray[1]) >= 73) {
```
matz3, line 74: Please use the `semver` package to compare versions. It's already a dependency of this project, so you just need to require it.

matz3 (changes requested):

lib/tasks/bundlers/generateFlexChangesBundle.js
MatthiasSchmalz, line 74: I'm just wondering if there is a version util class somewhere which does this

matz3: I'd suggest to use semver, as mentioned in my comment above.

thuyboehm: use semver

lib/tasks/bundlers/generateFlexChangesBundle.js
```diff
@@ -68,10 +69,10 @@ module.exports = async function({workspace, taskUtil, options: {namespace}}) {
 	const allResources = await workspace.byGlob(
 		`${pathPrefix}/changes/*.{change,variant,ctrl_variant,ctrl_variant_change,ctrl_variant_management_change}`);
 	if (allResources.length > 0) {
-		const version = await readManifestMinUI5Version();
-		const versionArray = version.split(".");
+		const MIN_UI5_VERION_1_73 = new Version("1.73.0");
+		const version = new Version(await readManifestMinUI5Version());
 		let hasFlexBundleVersion = false;
-		if (versionArray.length >= 2 && parseFloat(versionArray[0]) >= 1 && parseFloat(versionArray[1]) >= 73) {
+		if (version.major >= MIN_UI5_VERION_1_73.major && version.minor >= MIN_UI5_VERION_1_73.minor) {
```
MatthiasSchmalz: Could compare method be used?

thuyboehm: already use

MatthiasSchmalz: Ah I viewed an old version

matz3 (dismissed):

matz3:
@thuyboehm could you please sign the CLA via https://cla-assistant.io/SAP/ui5-builder?pullRequest=706 ?
It seems that the automatic comment hasn't been posted, and the check is currently pending.

lib/tasks/bundlers/generateFlexChangesBundle.js
```diff
@@ -68,9 +69,10 @@ module.exports = async function({workspace, taskUtil, options: {namespace}}) {
 	const allResources = await workspace.byGlob(
 		`${pathPrefix}/changes/*.{change,variant,ctrl_variant,ctrl_variant_change,ctrl_variant_management_change}`);
 	if (allResources.length > 0) {
-		const version = await readManifestMinUI5Version();
+		const MIN_UI5_VERION_1_73 = new Version("1.73.0");
+		const version = new Version(await readManifestMinUI5Version());
 		let hasFlexBundleVersion = false;
-		if (parseFloat(version) >= 1.73) {
+		if (version.compare(MIN_UI5_VERION_1_73) >= 0) {
```
RandomByte, line 75: Just my two cents: This is a pretty odd way of using the semver API, that I haven't seen so far and also couldn't find in the documentation at https://github.com/npm/node-semver#readme

IMHO `version.compare(version, "1.73.0") >= 0` would be sufficient and does not involve the creation of `Version` instances.

But I guess your solution works too and since it's your code we are totally fine with merging this as is. **Please let us know soon in case you want to change this. Otherwise we will proceed with releasing this PR.**

RandomByte: Edit: I meant `semver.compare`, not `version.compare`

thuyboehm: i just change it

codeworrior: Isn't it necessary to `semver.coerce` the version from the manifest first? I recently worked with `semver` for the first time and was puzzled that it doesn't do this automatically, at least not with the convenience comparison methods (`semver.lt` etc.).

matz3: Yes, it is. As the manifest.json schema doesn't validate the "minUI5Version" properly (just type "string"), it is likely that the patch version is missing in some cases (e.g. `1.100`). This should also be covered in the tests.

matz3: Well, even out `manifestCreator` just uses major.minor without the patch version.

codeworrior: ...which totally makes sense to me as the patch version doesn't play a role wrt. APIs. It's just that `semver` doesn't handle such short versions with a call to `semver.coerce`.

matz3:
Maybe some test cases for the "major.minor" version would be good to have, right?

thuyboehm:
It is ok now?

matz3 (approved):

matz3:
Released via [@ui5/builder v2.11.4](https://github.com/SAP/ui5-builder/releases/tag/v2.11.4) / [@ui5/cli v2.14.6](https://github.com/SAP/ui5-cli/releases/tag/v2.14.6)

PR #722, opened 2022-03-24
[FEAT]: enable build with CSS variables

flovogt:


lib/builder/ProjectBuildContext.js
```diff
@@ -37,6 +37,14 @@ class ProjectBuildContext {
 		return this._project === this._buildContext.getRootProject();
 	}
 
+	getOptions() {
+		return this._buildContext.getOptions();
```
flovogt, line 41: currently, `getOptions` is not used anywhere in our implementation. Also not in the build context. So I tend to remove them. @matz3 your opinion on that?

lib/tasks/TaskUtil.js
```diff
@@ -108,6 +108,28 @@ class TaskUtil {
...
+	getBuildOptions() {
+		return this._projectBuildContext.getOptions();
+	}
+
+	/**
+	 * Retrieves a build option defined by it's <code>key</code.
+	 * If no option with the given <code>key</code> is stored, <code>undefined</code> is returned.
+	 *
+	 * @param {string} key The option key
+	 * @returns {object} The build options
```
codeworrior, line 126: @returns {object|undefined} The build option (or undefined)

It's a single option here, not options.

lib/builder/ProjectBuildContext.js
matz3: Agree. Let's for now just add `getOption`.

flovogt: Done

lib/tasks/TaskUtil.js
flovogt: Thanks @codeworrior. Done

matz3:
PR should go into the `next` branch, not master.

PR #724, opened 2022-03-29
[INTERNAL] ApplicationBuilder: Align task order of "minify"

matz3:
In the LibraryBuilder the task is executed before the
generateManifestBundle and before any of the bundle tasks.
Therefore, the minify task should be executed right after the
replaceVersion task.


matz3:
To be checked with @RandomByte if this is the expected order or whether the LibraryBuilder should be changed.

codeworrior (approved):
LGTM

RandomByte (approved):
LGTM

PR #728, opened 2022-03-30
[FEATURE] builder: add cssVariables option

flovogt:
Add cssVariables option in the UI5-Builder.
BLI: CPOUI5FOUNDATION-346


lib/builder/ProjectBuildContext.js
```diff
@@ -39,6 +39,10 @@ class ProjectBuildContext {
 		return this._project === this._buildContext.getRootProject();
 	}
 
+	getOption(key) {
```
flovogt, line 42: Should we rename this function to `getBuildContextOption` or is this fine?

test/lib/types/themeLibrary/ThemeLibraryBuilder.js
```diff
@@ -44,6 +44,12 @@ test("tasks", async (t) => {
 		taskUtil: {
 			isRootProject: () => {
 				return true;
+			},
+			getBuildOptions: () => {
```
flovogt, line 48: can be removed

flovogt: Done

lib/tasks/TaskUtil.js
```diff
@@ -108,6 +108,18 @@ class TaskUtil {
 		return this._projectBuildContext.isRootProject();
 	}
 
+	/**
+	 * Retrieves a build option defined by it's <code>key</code.
+	 * If no option with the given <code>key</code> is stored, <code>undefined</code> is returned.
+	 *
+	 * @param {string} key The option key
+	 * @returns {object|undefined} The build option (or undefined)
```
matz3, line 116: The option value can be of any type, right?
```suggestion
	 * @returns {any|undefined} The build option (or undefined)
```

lib/builder/ProjectBuildContext.js
matz3: I think it's fine. It's anyway a private class.

lib/builder/builder.js
```diff
@@ -249,7 +250,12 @@ module.exports = {
 			virBasePath: "/"
 		});
 
-		const buildContext = new BuildContext({rootProject: tree});
+		const buildContext = new BuildContext({
+			rootProject: tree,
+			options: {
+				cssVariables: cssVariables
```
matz3, line 256: I wonder whether we should have some pre-fixes for the options key (similar to the tags), in case we want to allow a more generic approach for custom tasks.
But as we don't document the "cssVariables" option as API, we should be able to also do this later on, so IMO this is fine.

flovogt: So you mean create a `BuildOptionsCollection` like `ResourceTagCollection`?

lib/tasks/TaskUtil.js
flovogt: thats true. While introducing `BuildOptionsCollection` we could limit the kind of values

lib/tasks/TaskUtil.js
```diff
@@ -108,6 +108,18 @@ class TaskUtil {
 		return this._projectBuildContext.isRootProject();
 	}
 
+	/**
+	 * Retrieves a build option defined by it's <code>key</code.
```
KlattG, line 112:
```suggestion
	 * Retrieves a build option defined by its <code>key</code.
```

matz3 (dismissed):

KlattG (changes requested):

matz3 (approved):

PR #730, opened 2022-04-06
[FEATURE] generateThemeDesignerResources task: Create css_variables.less

flovogt:
[FEATURE] Enhance task generateThemeDesignerResources.js to create css_variables.less files.
BLI: CPOUI5FOUNDATION-487


flovogt:
Line https://github.com/SAP/ui5-builder/pull/730/files#diff-92c6913aec2b1a1b12c249189bafc7a0abbe5ccee607f72e847023916ed58796L59 and the following lines are creating a complete new `.theming` file. So, the existing theming file is ignored is this desired?

matz3:
Yes, the `.theming` file is either generated or copied from the `sap/ui/core` namespace. If it exists in the sources it will be overwritten.

lib/tasks/generateThemeDesignerResources.js
```diff
@@ -80,6 +153,39 @@ async function generateThemeDotTheming({workspace, combo, themeFolder}) {
...
+			cssVariablesSourceLessResource.map((cssVariableSourceLess) => {
+				const themeFolder = posixPath.dirname(cssVariableSourceLess.getPath());
+				log.verbose(`Generating css_variables.less for theme ${themeFolder}`);
+				return createCssVariablesLessResource({
+					workspace, combo, themeFolder
+				});
+			})
+		);
+
+		return cssVariablesLessResources.map(async (resource) => {
```
matz3, line 181: Isn't `Promise.all` missing here?

matz3 (changes requested):

lib/tasks/generateThemeDesignerResources.js
flovogt: Agreed on moving write logic to loop above

flovogt: Done

matz3 (approved):

PR #734, opened 2022-04-15
[FEATURE] Enable modern preload bundles without dependencies

RandomByte:
Currently, when building a UI5 project without dependencies, the legacy bundle format is used. This change defaults to the new bundle format and removes any checks for the existence of core modules from the bundle Builder.

* bundle/Builder: Remove checks for existence of core modules and
  replace them with bundleOptions "decorateBootstrapModule" and
  "legacyBundleFormat"
* generate(Component|Library)Preload: Stop globbing dependency resources

Tested with openui5-sample-app, which comes up fine after this change.

## ~Still open~:  
🥇 **Decision for now: Remove legacy API support completely. All 3.0 bundles use the new API.**

* Should we keep an option to still use the the legacy bundle format or remove it altogether?
* If yes, decide on when to set such a `legacyBundleFormat` parameter. Possible options:
    1. Based on detected framework version
    2. Based on CLI flag
        * This might be the most reasonable approach. We could also add a ui5.yaml setting to default this on root-project level
    3. Only based on bundleOption - this seems unreasonable since the root project might need to make this decision for all projects based on the target UI5 version
    4. Based on the minUI5Version from manifest.json
* Update tests based on decision


lib/processors/bundlers/moduleBundler.js
```diff
@@ -91,9 +91,9 @@ const log = require("@ui5/logger").getLogger("builder:processors:bundlers:module
  * @typedef {object} ModuleBundleOptions
  * @property {boolean} [optimize=true] Whether the module bundle gets minified
  * @property {boolean} [sourceMap=true] Whether to generate a source map file for the bundle
- * @property {boolean} [decorateBootstrapModule=false] If set to 'false', the module won't be decorated
+ * @property {boolean} [decorateBootstrapModule=false] If set to 'false', bootable bundles won't be decorated
```
matz3, line 94: This should then also be updated in the docs: https://sap.github.io/ui5-tooling/pages/Configuration/

matz3 (approved):

lib/processors/bundlers/moduleBundler.js
codeworrior: or, for now, in the v3 migration guide?

matz3: I think it's also applicable to the master, as only the text has been improved without changing the behavior, right?

RandomByte: Done in https://github.com/SAP/ui5-tooling/pull/616

RandomByte:
After force-pushing to this branch the PR didn't show the new commits. After closing I also couldn't reopen it. Maybe because some comments where referring directly to one of the old commits? 🤔 

Anyways, here's the same PR again: https://github.com/SAP/ui5-builder/pull/739

PR #735, opened 2022-04-20
[INTERNAL] generateLibraryManifest: Ignore dependency resources

RandomByte:
They should not be required to generate a manifest.json for a library
project.

### DIFF
--- a/lib/tasks/generateLibraryManifest.js
+++ b/lib/tasks/generateLibraryManifest.js
@@ -2,7 +2,6 @@
 
 const log = require("@ui5/logger").getLogger("builder:tasks:generateLibraryManifest");
 const manifestCreator = require("../processors/manifestCreator");
-const ReaderCollectionPrioritized = require("@ui5/fs").ReaderCollectionPrioritized;
 
 
 /**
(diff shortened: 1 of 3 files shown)


lib/tasks/generateLibraryManifest.js
```diff
@@ -12,23 +11,18 @@ const ReaderCollectionPrioritized = require("@ui5/fs").ReaderCollectionPrioritiz
  * @alias module:@ui5/builder.tasks.generateLibraryManifest
  * @param {object} parameters Parameters
  * @param {module:@ui5/fs.DuplexCollection} parameters.workspace DuplexCollection to read and write files
- * @param {module:@ui5/fs.AbstractReader} parameters.dependencies Reader or Collection to read dependency files
```
matz3, line 15: The parameter can also be removed within LibraryBuilder, right?

codeworrior:
Good catch! I guess I initially assumed that the dependency resources might be needed to determine the version of library dependencies. But already in the initial PR, those versions have been determined from the project's metadata, not from  resources in the dependencies.

lib/tasks/generateLibraryManifest.js
RandomByte: Done

matz3 (approved):

matz3 (approved):

PR #736, opened 2022-04-21
[FIX] generateResourcesJson: Add raw-module info for debug bundles

matz3:
The raw-module info within .library is only maintained using the
non-debug name. Therefore, the analysis must use that name when running
with debug bundles.

Furthermore the handling of the "module" property for debug resources
has been changed to only take place within the ResourceCollector.

JIRA: CPOUI5FOUNDATION-480

### DIFF
--- a/lib/lbt/resources/ResourceCollector.js
+++ b/lib/lbt/resources/ResourceCollector.js
@@ -286,6 +286,11 @@ class ResourceCollector {
 
 			// FIXME: "merged" property is only calculated in ResourceInfo#copyFrom
 			// Therefore using the same logic here to compute it.
+
+			// TODO: Idea: Use IsDebugVariant tag to decide whether to analyze the resource
+			// If the tag is set, we don't expect different analysis results so we can copy the info (else-path)
+			// Only when the tag is not set, we analyze the resource with its name (incl. -dbg)
+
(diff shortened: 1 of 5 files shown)


flovogt (commented):
LGTM

lib/lbt/resources/ResourceCollector.js
```diff
@@ -279,37 +279,43 @@ class ResourceCollector {
...
 				// We need to analyze the dbg resource if there is no non-dbg variant or
 				// it is a bundle because we will (usually) have different content.
-				debugBundlePromises.push(
-					this.enrichWithDependencyInfo(dbgInfo)
-				);
+
+				// In order to retrieve the correct raw module info, the name must be set
+				// to the non-debug name. The same happens during bundling, where also only
+				// the non-debug names are used.
+				dbgInfo.name = nonDebugName;
```
matz3, line 296: Changing the name here causes the ResourcePool to return the cached non-debug info instead of using that name to analyze the debug file.

So we need another solution here. Probably by checking for the IsDebugVariant tag to decide whether to analyze the resource.

matz3: This should be fixed now

RandomByte:
LGTM

flovogt:
LGTM

flovogt (approved):

codeworrior (approved):
LGTM

PR #741, opened 2022-05-02
[FIX] Avoid redundant bundle creation

matz3:
When a custom bundle has the same name as a built-in bundle (e.g.
foo/bar/library-preload.js), then the creation of the built-in bundle
can be skipped.

### DIFF
--- a/lib/tasks/bundlers/generateComponentPreload.js
+++ b/lib/tasks/bundlers/generateComponentPreload.js
@@ -21,10 +21,11 @@ const {negateFilters} = require("../../lbt/resources/ResourceFilterList");
  * inclusion overrides an earlier exclusion, and vice versa.
  * @param {string[]} [parameters.options.paths] Array of paths (or glob patterns) for component files
  * @param {string[]} [parameters.options.namespaces] Array of component namespaces
+ * @param {string[]} [parameters.options.customBundleNames] List of custom bundles defined for the project
  * @returns {Promise<undefined>} Promise resolving with <code>undefined</code> once data has been written
  */
 module.exports = function({
-	workspace, taskUtil, options: {projectName, paths, namespaces, excludes = []}
+	workspace, taskUtil, options: {projectName, paths, namespaces, customBundleNames = [], excludes = []}
 }) {
 	let nonDbgWorkspace = workspace;
 	if (taskUtil) {
@@ -65,6 +66,12 @@ module.exports = function({
 			const unusedFilterExcludes = new Set(allFilterExcludes);
 
 			const bundleDefinitions = allNamespaces.map((namespace) => {
+				const bundleName = `${namespace}/Component-preload.js`;
+				if (customBundleNames.includes(bundleName)) {
+					log.verbose(`skipping overridden built-in bundle ${bundleName}`);
+					return null;
+				}
+
 				const filters = [
 					`${namespace}/`,
 					`${namespace}/**/manifest.json`,
@@ -94,7 +101,7 @@ module.exports = function({
 				});
 
 				return {
-					name: `${namespace}/Component-preload.js`,
+					name: bundleName,
 					defaultFileTypes: [
 						".js",
 						".control.xml",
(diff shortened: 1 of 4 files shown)


lib/tasks/bundlers/generateComponentPreload.js
```diff
@@ -21,10 +21,11 @@ const {negateFilters} = require("../../lbt/resources/ResourceFilterList");
  * inclusion overrides an earlier exclusion, and vice versa.
  * @param {string[]} [parameters.options.paths] Array of paths (or glob patterns) for component files
  * @param {string[]} [parameters.options.namespaces] Array of component namespaces
+ * @param {string[]} [parameters.options.skipBundles] List of bundles to be skipped (i.e. not created)
```
RandomByte, line 24:
```suggestion
 * @param {string[]} [parameters.options.skipBundles] Names of bundles that should not be created
```

lib/tasks/bundlers/generateComponentPreload.js
```diff
@@ -65,6 +66,12 @@ module.exports = function({
 			const unusedFilterExcludes = new Set(allFilterExcludes);
 
 			const bundleDefinitions = allNamespaces.map((namespace) => {
+				const bundleName = `${namespace}/Component-preload.js`;
+				if (skipBundles.includes(bundleName)) {
+					log.verbose(`skipping overridden built-in bundle ${bundleName}`);
```
RandomByte, line 71:
```suggestion
					log.verbose(`Skipping generation of bundle ${bundleName} as requested`);
```

lib/tasks/bundlers/generateLibraryPreload.js
```diff
@@ -285,6 +244,14 @@ module.exports = function({workspace, taskUtil, options: {projectName, excludes
 		});
 	}
 
+	const execModuleBundlerIfNeeded = ({options, resources}) => {
+		if (skipBundles.includes(options.bundleDefinition.name)) {
+			log.verbose(`skipping overridden built-in bundle ${options.bundleDefinition.name}`);
```
RandomByte, line 249:
```suggestion
			log.verbose(`Skipping generation of bundle ${bundleName} as requested`);
```

lib/tasks/bundlers/generateLibraryPreload.js
```diff
@@ -274,9 +232,10 @@ function getSapUiCoreBunDef(name, filters, preload) {
  * inclusion overrides an earlier exclusion, and vice versa.
  * @param {object} parameters.options Options
  * @param {string} parameters.options.projectName Project name
+ * @param {string[]} [parameters.options.skipBundles] List of bundles to be skipped (i.e. not created)
```
RandomByte, line 235:
```suggestion
 * @param {string[]} [parameters.options.skipBundles] Names of bundles that should not be created
```

RandomByte (changes requested):

RandomByte (approved):

lib/tasks/bundlers/generateLibraryPreload.js
```diff
@@ -326,32 +293,32 @@ module.exports = function({workspace, taskUtil, options: {projectName, excludes
...
+					execModuleBundlerIfNeeded({
 						options: getModuleBundlerOptions({
 							name: "sap-ui-core-nojQuery-dbg.js", filters, preload: false, provided: true,
 							moduleNameMapping: unoptimizedModuleNameMapping
 						}),
 						resources: unoptimizedResources
 					}),
 				]).then((results) => {
-					const bundles = Array.prototype.concat.apply([], results);
+					const bundles = Array.prototype.concat.apply([], results).filter(Boolean);
```
codeworrior, line 321: Is the concat still needed? Promise.all always resolves with an array and the filter(Boolean) anyhow creates a new array.

codeworrior: Or is this meant to flatten the results ? `Array.prototype.flat` should be available since NodeJS 11 ...

matz3: Yes, it should flatten the results. We do this in several places, so maybe we should do some general code-cleanup change.

RandomByte: Yes, this is a common pattern in the UI5 Tooling code base. We should clean those things up in one sweep

codeworrior: okay with me.

PR #757, opened 2022-06-09
Update JSDoc template with current openui5/master

codeworrior:
Note: this PR contains two commits which might not be backward compatible (to be discussed)


RandomByte:
Rebased to resolve test issues

codeworrior:
The topic of backward compatibility of the template part has been discussed and we decided to keep the templates aligned between builder and openui5. Currently, the JSDoc build step of the tooling is mainly used for internal testing purposes and there alignment with the productively used openui5 template is more important than the ability to build JSDoc for older releases with the most recent ui5-tooling.

Plus, the compatibility issues mainly stems from necessary/missing changes in the SDK part and we are not aware of a scenario where the combination of ui5-tooling and SDK is used other than during SAP internal development of openui5/sapui5.

codeworrior:
ping

sebbi08:
Hi so from our side we would like the option for changing the JSDoc/ API.JSON template.

We are using the JSDoc build to create a demokit for our custom libraries. 

Currently, we are forced to stay at a specific ui5 cli version depending on the ui5 version the library supports.

RandomByte (approved):

RandomByte:
> Hi so from our side we would like the option for changing the JSDoc/ API.JSON template.
> 
> We are using the JSDoc build to create a demokit for our custom libraries.
> 
> Currently, we are forced to stay at a specific ui5 cli version depending on the ui5 version the library supports.

Thanks for the input! Yes, this is still a gap. I  guess this enhancement is still tracked here: https://github.com/SAP/ui5-tooling/issues/356

Unfortunately I still don't have a good idea on how to easily solve this. Maybe an option to provide your own plugin would be a good solution for the time being?

PR #767, opened 2022-07-14
[INTERNAL] Fix tests that broke with Terser v5.14.2

RandomByte:
Similar to https://github.com/SAP/ui5-builder/commit/4789c2079a460447171dbbf4627ae6d0b56f9e6c

Also exchange dependency "sourcemap-codec" with
"@jridgewell/sourcemap-codec", which is also used by Terser. This
reduces the overall number of dependencies. Supposedly this fork also
implements some improvments over the original module.


RandomByte:
Source maps seem to be more correct now:

**Before**
Screenshot 2022-07-14 at 17 10 03


**After**
Screenshot 2022-07-14 at 17 10 50

matz3 (approved):

PR #770, opened 2022-07-27
[FIX] minify: Change default to not omit source map resources

RandomByte:
Source map resources for non-bundle files are required to be part of the
build result if that is used as input for another build. Otherwise
custom bundles packaging those resources won't be able to find the
corresponding source maps, leading to basically unusable source maps.


codeworrior (commented):
LGTM, but I'll give @matz3 a chance for a review

matz3 (approved):

PR #775, opened 2022-08-04
[FEATURE] jsdoc: Support destructuring of enums for defaultValue

d3xter666:
JIRA: CPOUI5FOUNDATION-503

This PULL request tries to enable ES6+ destructuring for UI5 enums.

There might be different cases and ways to achieve this and with this change, we try to cover the ones we already know. (Note: There might be more we're not aware of).

Here are the cases we try to cover:

1. Destructuring from fully quantified name
```javascript
// shortcut for sap.ui.core.TitleLevel
var {TitleLevel} = sap.ui.core;
```

1. Destructuring from fully quantified name
```javascript
// shortcut for sap.ui.core.TitleLevel
var {TitleLevel} = sap.ui.core;
```

2. Chained destructuring
```javascript
const library = {
  MyEnum: {
    Foo: "FooValue",
  },
};


const { MyEnum } = library;
const { Foo } = MyEnum;
```

3. Destructuring with a rename
```javascript
 const { Foo: Bar } = MyEnum; 
```

4. Destructuring as an argument of an arrow function
```javascript
sap.ui.define([
	"sap/ui/core/Control", 
	"./library"
], (Control, { MyEnumFromLib } ) => {
...
});
```

5. Not a destructuring, but a pattern close to # 3 and already covered by the builder. However, we need to be aware of it and should consider it as the chain in AST is (almost) the same as the one in # 3
```javascript
var MyTitleLevelVar = sap.ui.core.TitleLevel;
```

6. Any mixtures from the cases above


lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -516,6 +523,192 @@ function isPotentialEnum(node) {
...
+		.filter((curEnumKey) => {
+			const [enums, resource] = curEnumKey.split("||");
+
+			// Check the resource
+			if (resource.indexOf(localModule) === -1) {
+				return false;
+			}
+
+			// Check if the key is in the enums
+			return enums.split("|").includes(potentialEnumKey);
```
d3xter666, line 552: Not sure whether this is the correct way to check this as this filtering relies on ENUM's:
1. Filename of where it's been defined (this one is ok)
2. Check whether any of the enums in the file contains a such element.

It should be fine as the chain it creates is:
{dependencyName}.{enumName}.{enumElement} i.e. library.XXX.Foo
Foo could be defined in other enums places in the same file, but the check is just whether it's there or not. The XXX (eunm's name) is derived in the previous step by traversing up to the tree in the previous step.

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -646,6 +835,11 @@ function convertValueWithRaw(node, type, propertyName) {
 				raw: local.raw
 			};
 		}
+
+		const potentialEnum = resolvePotemtialEnum(node, type);
```
d3xter666, line 839: This is the case where the ENUM could be just the element. For example, `const { TitleLevel } = sap.ui.core;`

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -516,6 +523,192 @@ function isPotentialEnum(node) {
...
+ * Gets enclosing scope
+ * 
+ * @param {Node} node 
+ * @returns {Scope}
+ */
+ function getEnclosingVariableScope (node) {
+	// Get to the nearest upper scope
+	let nearestScopeableNode = node;
+	while (
+		nearestScopeableNode && nearestScopeableNode.type &&
```
d3xter666, line 699: Probably not the full list as for example `let` and `const` are block-scoped, but for some reason I'm not able to derive the scope in such cases. Probably would need to check if this is an issue or missing functionality in `escope`

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -516,6 +523,190 @@ function isPotentialEnum(node) {
...
+		nearestScopeableNode.type !== Syntax.ArrowFunctionExpression &&
+		nearestScopeableNode.type !== Syntax.FunctionDeclaration &&
+		nearestScopeableNode.type !== Syntax.FunctionExpression
+	) {
+		nearestScopeableNode = nearestScopeableNode.parent;
+	}
+
+	const scopeManager = escope.analyze(currentProgram);
```
codeworrior, line 705: This analyse step ideally should be done only once. Either lazy on first access or when the currentProgram is set.

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -516,6 +523,190 @@ function isPotentialEnum(node) {
...
+ function getEnclosingVariableScope (node) {
+	// Get to the nearest upper scope
+	let nearestScopeableNode = node;
+	while (
+		nearestScopeableNode && nearestScopeableNode.type &&
+		nearestScopeableNode.type !== Syntax.ArrowFunctionExpression &&
+		nearestScopeableNode.type !== Syntax.FunctionDeclaration &&
+		nearestScopeableNode.type !== Syntax.FunctionExpression
+	) {
+		nearestScopeableNode = nearestScopeableNode.parent;
```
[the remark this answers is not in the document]
d3xter666: I'd double-check it, but as far as I got it, it comes from the `escope`

lib/processors/jsdoc/lib/ui5/plugin.js
d3xter666: Now it should be ok. Setting it up where the currentProgram is being set

lib/processors/jsdoc/lib/ui5/plugin.js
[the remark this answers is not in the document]
d3xter666: So, it's indeed from `escope`. Here's the definition: https://github.com/estools/escope/blob/master/src/definition.js#L47

lib/processors/jsdoc/lib/ui5/plugin.js
d3xter666: Now should be better : )
Thanks a lot!

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -516,6 +526,190 @@ function isPotentialEnum(node) {
...
+ * @returns {Object}
+ */
+function resolvePotemtialEnum(node, type) {
+	const varsDestructuringStack = resolveFullyQunatifiedName(node, type);
+
+	if (!varsDestructuringStack || !varsDestructuringStack.length) {
+		return null;
+	}
+
+	// When the fully qunatified name gets resolved, we could try the "old" approach by checking
```
flovogt, line 544: quantified

flovogt (changes requested):

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -646,6 +836,11 @@ function convertValueWithRaw(node, type, propertyName) {
 				raw: local.raw
 			};
 		}
+
+		const potentialEnum = resolvePotemtialEnum(node, type);
```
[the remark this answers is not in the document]
d3xter666: Thanks!

lib/processors/jsdoc/lib/ui5/plugin.js
d3xter666: Thanks!

test/expected/build/library.j/dest/resources/library/j/some.js
```diff
@@ -1,16 +1,84 @@
...
+						type: "testlib.MyValidEnum",
+						group: "Misc",
+						defaultValue: MyValidEnum.Foo,
+					},
+
+					/**
+					 * validPropertyDefaultValueEnumViaDestructuringGlobal
+					 */
+					validPropertyDefaultValueEnumViaDestructuringGlobal: {
+						type: "sap.ui.core",
```
matz3, line 73: This doesn't sound like a valid type...

test/expected/build/library.j/dest/resources/library/j/some.js
```diff
@@ -1,16 +1,84 @@
...
+						group: "Misc",
+						defaultValue: TitleLevel,
```
matz3, line 75: ... and also the default value doesn't make sense. Or is this a test for some invalid use case?

matz3 (changes requested):

test/expected/build/library.j/dest/resources/library/j/some.js
d3xter666: Indeed :D 
Now should be the correct one

test/expected/build/library.j/dest/resources/library/j/some.js
d3xter666: Nope, it was a mistake. Now should be the correct one

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -249,13 +253,19 @@ function analyzeModuleDefinition(node) {
...
+
 		currentModule.factory = args[arg];
 		arg++;
 	}
 	if ( currentModule.dependencies && currentModule.factory ) {
 		for ( let i = 0; i < currentModule.dependencies.length && i < currentModule.factory.params.length; i++ ) {
-			const name = currentModule.factory.params[i].name;
+			const name =
+				(currentModule.factory.params[i].type === Syntax.ObjectPattern)
+					? currentModule.factory.params[i].properties[0].value.name // ObjectPattern means destructuring of the parameter
```
matz3, line 266: What happens if one is destructuring more than one property?
If I understand this code correctly, it will only pick the first one, right?
```js
function({ Foo, Bar }) {}
```

matz3 (changes requested):
I only briefly looked through the code, but will need to run it locally to fully verify the resolve logic

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -516,6 +526,190 @@ function isPotentialEnum(node) {
...
+
+	if (!varsDestructuringStack || !varsDestructuringStack.length) {
+		return null;
+	}
+
+	// When the fully quantified name gets resolved, we could try the "old" approach by checking 
+	// the type + key
+	let value = varsDestructuringStack.join(".");
+	
+	if ( value.indexOf(type + ".") === 0 ) {
```
matz3, line 548: you might also use `startsWith`

d3xter666: Done

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -274,15 +274,17 @@ function analyzeModuleDefinition(node) {
...
 			names.forEach(name => {
 				const module = resolveModuleName(currentModule.module, currentModule.dependencies[i]);
-				debug(`  import ${name} from '${module}'`);
-				currentModule.localNames[name] = {
-					module: module
+				debug(`  import ${name.renamed || name.original} from '${module}'`);
+
+				currentModule.localNames[name.renamed || name.original] = {
+					module: module,
+					property: name.original,
```
matz3, line 287: This has been enhanced but I don't see any usage of the new property. How is the rename/descructuring use case actually solved now?

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -1009,11 +1012,15 @@ function convertValueWithRaw(node, type, propertyName) {
...
 		}
-
+		
 	} else if ( isTemplateLiteralWithoutExpression(node) ) {
+		let value = {};
+		if (node && node.quasis && node.quasis[0] && node.quasis[0].value) {
+			value = node.quasis[0].value;
+		}
 		return {
-			value: node?.quasis?.[0]?.value?.cooked,
```
matz3, line 805: I think here we can expect that `node.quasis[0].value` is defined because `isTemplateLiteralWithoutExpression` already checks for that.

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -803,8 +806,8 @@ function resolveFullyQuantifiedName(node, type) {
 
 		if (potentialChunk && potentialChunk.renamed) {
 			originalName = originalName.replace(
-				potentialChunk.renamed + ".",
-				potentialChunk.original + "."
+				new RegExp("(^|.)" + potentialChunk.renamed + "($|.)"),
```
matz3, line 809: Is there a need to escape characters for the RegExp?

matz3 (changes requested):

lib/processors/jsdoc/lib/ui5/plugin.js
d3xter666: Thanks! Fixed it

lib/processors/jsdoc/lib/ui5/plugin.js
d3xter666: Replaced the Regex with a string comparison function

lib/processors/jsdoc/lib/ui5/plugin.js
d3xter666: To be honest, I don't exactly remember clearly the reasoning for resolving the variable to its original name. Still, I remember when we decided to keep the name as it had been defined in the code but record the original name as a property for traceability.

I just checked the file history and I believe this change should be the key here: https://github.com/SAP/ui5-builder/commit/cb2e115b2da034f030a2d3fe4774e6a7e2c3cac1

In the beginning, [this change](https://github.com/SAP/ui5-builder/commit/cb2e115b2da034f030a2d3fe4774e6a7e2c3cac1) was resolving properly just local variables, but not destructured (and renamed) function arguments and as far as I remember, that's why I needed to keep the original name of the restructured argument. From the current perspective, this now seems like it was a bug :D 
With the refactoring of the `checkVarRenaming()` and more specifically with [this reducer](https://github.com/SAP/ui5-builder/pull/775/files#diff-693e706995c87056125c1cff2fe8953f9592038c8f50206171a84a6d2244d8ddR701), it is now able to correctly resolve destructuring and renaming of arguments

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -642,6 +657,207 @@ function isPotentialEnum(node) {
...
+/**
+ * Gets enclosing scope
+ * 
+ * @param {Node} node 
+ * @returns {Scope}
+ */
+ function getEnclosingVariableScope (node) {
+	// Get to the nearest upper scope
+	let nearestScopeableNode = node;
+	let neatestScope = scopeManager.acquire(nearestScopeableNode);
```
matz3, line 847: Should be `nearestScope`, right?

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -642,6 +657,207 @@ function isPotentialEnum(node) {
...
+	// We would eiter ways would resolve to the same leftmost token.
+	let leftMostName = getLeftmostName(node);
+	let originalName = getObjectName(node) || "";
+	const currentScope = getEnclosingVariableScope(node);
+
+	if (!currentScope) {
+		return null;
+	}
+
+	// TODO: Check for hierarchical destructuring
```
matz3, line 763: Is this still a TODO?

matz3 (changes requested):

lib/processors/jsdoc/lib/ui5/plugin.js
d3xter666: Yep :D

lib/processors/jsdoc/lib/ui5/plugin.js
d3xter666: It's obsolete indeed :)

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -733,18 +946,15 @@ function convertValueWithRaw(node, type, propertyName) {
...
-//		} else if ( value.indexOf(type.split(".").slice(-1)[0] + ".") === 0 ) {
-//			// unqualified name might be a local name (just a guess - would need static code analysis for proper solution)
-//			return value.slice(type.split(".").slice(-1)[0].length + 1);
+		const potentialEnum = resolvePotentialEnum(node, type);
+
+		if ( potentialEnum ) {
+			return potentialEnum;
+			//		} else if ( value.indexOf(type.split(".").slice(-1)[0] + ".") === 0 ) {
+			//			// unqualified name might be a local name (just a guess - would need static code analysis for proper solution)
+			//			return value.slice(type.split(".").slice(-1)[0].length + 1);
```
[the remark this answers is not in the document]
d3xter666: Exactly

test/expected/build/library.j/dest/resources/library/j/some.js
```diff
@@ -1,20 +1,106 @@
...
+	 *
+	 * @author SAP SE
+	 * @version ${version}
+	 *
+	 * @public
+	 * @alias testlib.ValidPropertyDefaultValue
+	 * @ui5-metamodel text
+	 */
+	var ValidPropertyDefaultValue = Control.extend(
+		"testlib.ValidPropertyDefaultValue",
```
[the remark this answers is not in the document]
d3xter666: Thanks

test/expected/build/library.j/dest/resources/library/j/dependency2.js
```diff
@@ -0,0 +1,34 @@
+/*!
+ * ${copyright}
+ */
+sap.ui.define([], function () {
+	sap.ui.getCore().initLibrary({
+		name: "testlib",
```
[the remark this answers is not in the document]
d3xter666: Thanks! Now should be better :)

test/fixtures/library.j/main/src/library/j/dependency.js
```diff
@@ -0,0 +1,53 @@
+/*!
+ * ${copyright}
+ */
+sap.ui.define([], function () {
+	sap.ui.getCore().initLibrary({
+		name: "testlib",
```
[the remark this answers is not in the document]
d3xter666: Hopefully, now should be a bit better. I thought that, although those libraries are not truly needed, it might be 
 for good to have them with the ENUMs defined for the sake of clarity, readability, and for a better understanding of the code. The scenario with _sap.ui.core.libary_ and _sap.m.library_ inside a control, for example, seems to be a valid one. Though you're quite correct, and there isn't a place in UI5 where we have multiple library.js files defined in a single "library"

test/expected/build/library.j/dest/resources/library/j/ES6.features.2.js
```diff
@@ -21,14 +21,13 @@ window.someRandomModule ||
...
 		 * @public
 		 * @since 1.22
-		 * @alias library.j.aaa
-		 * @ui5-metamodel This control will also be described in the UI5 (legacy) design time meta model.
+		 * @alias library.j.SubControl
 		 */
-		(a) =>
-			a.extend(`library.j.aaa`, {
+		(Control) =>
+			Control.extend(`library.j.SubControl`, {
```
matz3, line 30: Still, the control name doesn't fit to the name of the file / module (`library/j/SubControl` vs `library/j/ES6.features.2`).

test/expected/build/library.j/dest/resources/library/j/ES6.features.2.js
```diff
@@ -21,14 +21,13 @@ window.someRandomModule ||
 		 * @version ${version}
 		 *
 		 * @constructor
-		 * @extends library.j.a
+		 * @extends library.j.Control
```
matz3, line 24: This doesn't exist. I'd rather suggest to just import and extend from sap.ui.core.Control.

test/expected/build/library.j/dest/resources/library/j/library-core.js
```diff
@@ -0,0 +1,34 @@
+/*!
```
matz3, line 1: In order to have the correct library namespace reflected, this file should be named `library/j/core/library.js`.

test/expected/build/library.j/dest/resources/library/j/library.js
```diff
@@ -0,0 +1,53 @@
+/*!
+ * ${copyright}
+ */
+sap.ui.define([], function () {
+	sap.ui.getCore().initLibrary({
+		name: "library.j",
+		version: "${version}",
+		dependencies: ["sap.ui.core"],
+		designtime: "library.j/designtime/library.designtime",
```
matz3, line 9: library/j

test/expected/build/library.j/dest/resources/library/j/some.js
```diff
@@ -1,20 +1,103 @@
...
+	 *
+	 * @author SAP SE
+	 * @version ${version}
+	 *
+	 * @public
+	 * @alias library.j.ValidPropertyDefaultValue
+	 * @ui5-metamodel text
+	 */
+	var ValidPropertyDefaultValue = Control.extend(
+		"library.j.ValidPropertyDefaultValue",
```
matz3, line 30: Control name doesn't fit to the name of the file / module (`library/j/ValidPropertyDefaultValue ` vs `library/j/some.js`).

matz3 (changes requested):

test/expected/build/library.j/dest/resources/library/j/ES6.features.2.js
d3xter666: Done

test/expected/build/library.j/dest/resources/library/j/ES6.features.2.js
d3xter666: Done

test/expected/build/library.j/dest/resources/library/j/library-core.js
d3xter666: Done

test/expected/build/library.j/dest/resources/library/j/some.js
d3xter666: Done

matz3 (approved):

PR #780, opened 2022-08-11
[FEATURE] Minifier: Support input source maps

RandomByte:
If a resource references a source map:
1. Use that source map for the debug variant of the resource
2. Pass it to terser so it can update it based on the transformation, 
   preserving the mapping to the initial source

The general logic is very similar to what we are already doing in [`lbt/bundle/Builder.js`](https://github.com/SAP/ui5-builder/blob/39b2b381d2014e1cc3b38fa69ce02b95be356121/lib/lbt/bundle/Builder.js#L274).

Open points (also see comments below):
* Should we use existing source maps by default?
    * If resources have been modified in the meantime (for example by preceding tasks such as `replaceVersion` or `replaceCopyright`), the corresponding source map is likely invalid and should not be used. It's better to have no map than an incorrect one
    * Until we find a solution to update source maps in tasks modifying resource content, we should probably make this opt-in. Maybe even limiting it to resources matching a provided glob pattern
    * ✅ **Result:** We agreed to check whether a resource has been modified during the UI5 Tooling build. If that is the case, we assume that the corresponding source map is potentially corrupt and ignore it.

JIRA: CPOUI5FOUNDATION-467


petermuessig:
@RandomByte instead of just relying on inline source maps, I would extend to logic a bit more and depend on the sourceMappingURL information, e.g. `lib/processors/minifier.js`:

```js
import {fileURLToPath} from "node:url";
import posixPath from "node:path/posix";
import os from "node:os";
import workerpool from "workerpool";
import Resource from "@ui5/fs/Resource";
import {getLogger} from "@ui5/logger";
const log = getLogger("builder:processors:minifier");

const debugFileRegex = /((?:\.view|\.fragment|\.controller|\.designtime|\.support)?\.js(?:\.map)?)$/;

const MIN_WORKERS = 2;
const MAX_WORKERS = 4;
const osCpus = os.cpus().length || 1;
const maxWorkers = Math.max(Math.min(osCpus - 1, MAX_WORKERS), MIN_WORKERS);

// Shared workerpool across all executions until the taskUtil cleanup is triggered
let pool;

function getPool(taskUtil) {
	if (!pool) {
		log.verbose(`Creating workerpool with up to ${maxWorkers} workers (available CPU cores: ${osCpus})`);
		const workerPath = fileURLToPath(new URL("./minifierWorker.js", import.meta.url));
		pool = workerpool.pool(workerPath, {
			workerType: "auto",
			maxWorkers
		});
		taskUtil.registerCleanupTask(() => {
			log.verbose(`Terminating workerpool`);
			const poolToBeTerminated = pool;
			pool = null;
			poolToBeTerminated.terminate();
		});
	}
	return pool;
}

async function minifyInWorker(options, taskUtil) {
	return getPool(taskUtil).exec("execMinification", [options]);
}

/**
 * @public
 * @module @ui5/builder/processors/minifier
 */

/**
 * Result set
 *
 * @public
 * @typedef {object} MinifierResult
 * @property {@ui5/fs/Resource} resource Minified resource
 * @property {@ui5/fs/Resource} dbgResource Debug (non-minified) variant
 * @property {@ui5/fs/Resource} sourceMap Source Map
 */

/**
 * Minifies the supplied resources.
 *
 * @public
 * @function default
 * @static
 *
 * @param {@ui5/fs/Resource[]} parameters.resources List of resources to be processed
 * @param {@ui5/builder/tasks/TaskUtil|object} [parameters.taskUtil] TaskUtil instance.
 *    Required when using the <code>useWorkers</code> option
 * @param {object} [parameters.options] Options
 * @param {boolean} [parameters.options.addSourceMappingUrl=true]
 *   Whether to add a sourceMappingURL reference to the end of the minified resource
 * @param {boolean} [parameters.options.useWorkers=false]
 *  Whether to offload the minification task onto separate CPU threads. This often speeds up the build process
 * @returns {Promise<module:@ui5/builder/processors/minifier~MinifierResult[]>}
 *   Promise resolving with object of resource, dbgResource and sourceMap
 */
export default async function({resources, taskUtil, options: {addSourceMappingUrl = true, useWorkers = false} = {}, workspace}) {
	let minify;
	if (useWorkers) {
		if (!taskUtil) {
			// TaskUtil is required for worker support
			throw new Error(`Minifier: Option 'useWorkers' requires a taskUtil instance to be provided`);
		}
		minify = minifyInWorker;
	} else {
		// Do not use workerpool
		minify = (await import("./minifierWorker.js")).default;
	}

	return Promise.all(resources.map(async (resource) => {
		const dbgPath = resource.getPath().replace(debugFileRegex, "-dbg$1");
		const dbgResource = await resource.clone();
		dbgResource.setPath(dbgPath);

		const filename = posixPath.basename(resource.getPath());
		const code = await resource.getString();

		const sourceMapOptions = {
			filename
		};
		if (addSourceMappingUrl) {
			sourceMapOptions.url = filename + ".map";
		}

		const dbgFilename = posixPath.basename(dbgPath);

		// find existing source map (either from sourceMappingURL or by naming convention)
		const sourceMappingURL = code.match(/^\/\/# sourceMappingURL=(.*)$/m)?.[1] || filename + ".map";
		if (sourceMappingURL.startsWith("data:")) {
			sourceMapOptions.content = "inline";
		} else {
			const sourceMap = await workspace.byPath(posixPath.resolve(posixPath.dirname(resource.getPath()), sourceMappingURL));
			if (sourceMap) {
				sourceMapOptions.content = await sourceMap.getString();
				const dbgSourceMapPath = sourceMap.getPath().replace(debugFileRegex, "-dbg$1");
				const dbgSourceMap = await sourceMap.clone();
				dbgSourceMap.setPath(dbgSourceMapPath);
				const dbgSourceMapJSON = JSON.parse(sourceMapOptions.content);
				dbgSourceMapJSON.file = dbgFilename;
				dbgSourceMap.setString(JSON.stringify(dbgSourceMapJSON));
				await workspace.write(dbgSourceMap);
			}
		}

		const result = await minify({
			filename,
			dbgFilename,
			code,
			sourceMapOptions
		}, taskUtil);
		resource.setString(result.code);
		const sourceMapResource = new Resource({
			path: resource.getPath() + ".map",
			string: result.map
		});
		return {resource, dbgResource, sourceMapResource};
	}));
}

```

I used this to verify the source maps handling in the `ui5-ecosystem-showcase` project by patching the `ui5-builder` with the above code (it also requires to pass the `workspace` so far to read the file defined in the `sourceMappingURL`. It works with inline source maps, with externals and it also creates the source maps for the `-dbg` files which maybe also are based on TypeScript or transpiled. Maybe we can chat about this soon...

RandomByte:
Thanks Peter. Looks good! Maybe we can also reuse the code from our [bundle/Builder module](https://github.com/SAP/ui5-builder/blob/5fbe1618f458b374f748412af9a1abd37a44b561/lib/lbt/bundle/Builder.js#L559)

The main question for me is whether we should always search for source maps in the code or rather make this a `ui5.yaml` configuration option. For this we should benchmark the performance impact of always doing it.

petermuessig:
@RandomByte @matz3 any chance to fix this before the vacation starts? 😉

RandomByte:
Some things we need to consider when using exiting source maps:

Existing source maps for resources that are modified by tasks preceding the minify tasks are likely to become inconsistent. For example if a resource is transpiled from TypeScript to JavaScript right at the beginning of the build, the `replaceCopyright` task might insert a multi-line copyright statement at the beginning of the JavaScript resource, thus invalidating the source map.

This is not an issue for current TypeScript scenarios where typically [`ui5-tooling-transpile`](https://www.npmjs.com/package/ui5-tooling-transpile) is being used and configured to run `afterTask: replaceVersion`. However, those projects then can't use the standard `replaceCopyright` or `replaceVersion` tasks and need to find other solutions for those.

To solve this aspect we would probably need to enhance the `Resource` class to maintain relationships between resources and their source map resources. And we need a central utility module to easily align source maps with the changes done by a task. Rollup and other bundlers have similar functionality, so it makes sense to integrate this into UI5 Tooling at some point.

petermuessig:
Any chance for a release? 😄

petermuessig:
Cc @matz3 @flovogt @d3xter666

lib/processors/minifier.js
```diff
@@ -38,6 +42,76 @@ async function minifyInWorker(options, taskUtil) {
...
+					} else {
+						throw new Error(`Not found: ${sourceMapPath}`);
+					}
+				} catch (e) {
+					// No input source map
+					log.warn(`Unable to read source map for resource ${resourcePath}: ${e.message}`);
+				}
+			}
+		} else {
+			const sourceMapFileCandidate = resourcePath + ".map";
```
matz3, line 99: We decided to remove this path, as it's not expected to have a Source Map file without a reference in the JavaScript file. Furthermore, if this is the case, it might be on purpose so that the existing Source Map is not used as input Source Map.

codeworrior: For the records: according to the source map specs, a server can send an HTTP header to provide a reference to the source map for a resource it delivers. This is the only case where a source map would be needed without a reference in the code. 

In the context of UI5, however, we're not aware of any scenario that makes use of this feature and therefore currently don't consider it as a use case for the tooling.

RandomByte:
Some benchmarks according to our [documentation](https://sap.github.io/ui5-tooling/v3/pages/Benchmarking/), comparing this PR to the current main branch:

## openui5-sample-app
ui5-builder Ref | Command | Mean [ms] | Min [ms] | Max [ms] | Relative
|:---|:---|---:|---:|---:|---:|
| main ([`e96a2be`](https://github.com/SAP/ui5-builder/commit/e96a2be8dd3cecfe94209e8d01bf651bc79edc3b)) | `ui5 build` | 730.3 ± 25.7 | 699.0 | 786.5 | Baseline
| minifier-input-sourcemap ([`b50869d`](https://github.com/SAP/ui5-builder/pull/780/commits/b50869d5527f5e2a819c4db0f70430a2c3d43072) | `ui5 build` | 714.3 ± 9.1 | 703.5 | 728.4 | -2% (basically the same)

## sap.ui.core

ui5-builder Ref | Command | Mean [s] | Min [s] | Max [s] | Relative
|:---|:---|---:|---:|---:|---:|
| main ([`e96a2be`](https://github.com/SAP/ui5-builder/commit/e96a2be8dd3cecfe94209e8d01bf651bc79edc3b)) | `ui5 build` | 8.757 ± 0.167 | 8.496 | 9.061 | Baseline
| minifier-input-sourcemap ([`b50869d`](https://github.com/SAP/ui5-builder/pull/780/commits/b50869d5527f5e2a819c4db0f70430a2c3d43072) | `ui5 build` | 8.600 ± 0.151 | 8.325 | 8.745 | -1.7% (basically the same)

## OpenUI5 testsuite (including dependencies)

ui5-builder Ref | Command | Mean [s] | Min [s] | Max [s] | Relative
|:---|:---|---:|---:|---:|---:|
| main ([`e96a2be`](https://github.com/SAP/ui5-builder/commit/e96a2be8dd3cecfe94209e8d01bf651bc79edc3b)) | `ui5 build --all` | 58.487 ± 0.541 | 57.440 | 59.186 | Baseline
| minifier-input-sourcemap ([`b50869d`](https://github.com/SAP/ui5-builder/pull/780/commits/b50869d5527f5e2a819c4db0f70430a2c3d43072) | `ui5 build --all` | 58.622 ± 1.507 | 56.833 | 60.867 | -0.8% (the same!)

## Internal TS Library

ui5-builder Ref | Command | Mean [s] | Min [s] | Max [s] | Relative
|:---|:---|---:|---:|---:|---:|
| main ([`e96a2be`](https://github.com/SAP/ui5-builder/commit/e96a2be8dd3cecfe94209e8d01bf651bc79edc3b)) | `ui5 build` | 5.528 ± 0.228 | 5.348 | 5.985 | Baseline
| minifier-input-sourcemap ([`b50869d`](https://github.com/SAP/ui5-builder/pull/780/commits/b50869d5527f5e2a819c4db0f70430a2c3d43072) | `ui5 build` | 5.520 ± 0.227 | 5.364 | 5.988 | The same


*Specs: MacBook Pro M1 Max*

lib/tasks/minify.js
```diff
@@ -19,32 +20,43 @@ import minifier from "../processors/minifier.js";
  * @param {string} parameters.options.pattern Pattern to locate the files to be processed
  * @param {boolean} [parameters.options.omitSourceMapResources=false] Whether source map resources shall
  * 		be tagged as "OmitFromBuildResult" and no sourceMappingURL shall be added to the minified resource
+ * @param {boolean} [parameters.options.useInputSourceMaps=false] Whether
```
matz3, line 23: parameter description incomplete

lib/processors/minifier.js
```diff
@@ -86,20 +160,69 @@ export default async function({resources, taskUtil, options: {addSourceMappingUr
...
+		const sourceMappingUrl = await extractAndRemoveSourceMappingUrl(resource);
+
+		const code = await resource.getString();
+		// Create debug variant based off the original resource before minification
+		const dbgResource = await resource.clone();
+		dbgResource.setPath(dbgPath);
+
+		let dbgSourceMapResource;
+		if (sourceMappingUrl) {
+			if (resource.getSourceMetadata()?.contentModified) {
```
matz3, line 186: This flag becomes `true` via `resource.setString` in `extractAndRemoveSourceMappingUrl`, which doesn't fit to what this check is supposed to do.

This doesn't appear in the current tests as the resources don't have sourceMetadata, thus the flag is not set at all. This applies to resources created by tasks, e.g. the transpile task.

But when files are already part of the project sources (i.e. transpiled upfront outside of ui5 build), the `contentModified` flag will always be `true`, so source maps will be ignored.

matz3 (dismissed):

lib/processors/minifier.js
matz3: Fixed via https://github.com/SAP/ui5-builder/pull/780/commits/c87f8834f3076db90f8156a09a1757f3eba82f05

petermuessig:
What's the status on this one here?

matz3:
> What's the status on this one here?

We've clarified all open points and are working on finalising this PR in the upcoming sprint.

RandomByte:
ui5-fs Resource change to ensure we don't use source maps of resources that have been modified by previous tasks, even when they exist in-memory only: https://github.com/SAP/ui5-fs/pull/524

lib/tasks/minify.js
RandomByte: Done

RandomByte:
Verified successful with all planned scenarios: `easy-ui5 ts-app`, `easy-ui5 ts-library` and an internal TS library.

This PR is ready from my side and only requires final review 👍

lib/processors/minifier.js
```diff
@@ -62,18 +123,31 @@ async function minifyInWorker(options, taskUtil) {
...
  * @param {@ui5/fs/Resource[]} parameters.resources List of resources to be processed
+ * @param {fs|module:@ui5/fs/fsInterface} parameters.fs Node fs or custom
+ *    [fs interface]{@link module:@ui5/fs/fsInterface}. Required when setting "readSourceMappingUrl" to true
  * @param {@ui5/builder/tasks/TaskUtil|object} [parameters.taskUtil] TaskUtil instance.
  *    Required when using the <code>useWorkers</code> option
  * @param {object} [parameters.options] Options
+ * @param {boolean} [parameters.options.readSourceMappingUrl=false]
+ *   Whether to make use of any existing source maps referenced in the resources to be minified. Use this to preserve
+ *   the reference to the original source files, such as TypeScript files, in the generated source map.<br>
+ *   If a resource has been modified by a previous tasks, any existing source map will be ignored regardless of this
```
KlattG, line 134:
```suggestion
 *   If a resource has been modified by a previous task, any existing source map will be ignored regardless of this
```

lib/processors/minifier.js
```diff
@@ -62,18 +123,31 @@ async function minifyInWorker(options, taskUtil) {
  *
  * @param {object} parameters Parameters
  * @param {@ui5/fs/Resource[]} parameters.resources List of resources to be processed
+ * @param {fs|module:@ui5/fs/fsInterface} parameters.fs Node fs or custom
+ *    [fs interface]{@link module:@ui5/fs/fsInterface}. Required when setting "readSourceMappingUrl" to true
  * @param {@ui5/builder/tasks/TaskUtil|object} [parameters.taskUtil] TaskUtil instance.
  *    Required when using the <code>useWorkers</code> option
  * @param {object} [parameters.options] Options
+ * @param {boolean} [parameters.options.readSourceMappingUrl=false]
+ *   Whether to make use of any existing source maps referenced in the resources to be minified. Use this to preserve
```
KlattG, line 132:
```suggestion
 *   Whether to make use of any existing source maps referenced in the resources to be minified. Use this option to preserve
```

KlattG (changes requested):
Just some minor stuff - hope I understood it correctly.

lib/tasks/minify.js
```diff
@@ -19,32 +20,45 @@ import minifier from "../processors/minifier.js";
  * @param {string} parameters.options.pattern Pattern to locate the files to be processed
  * @param {boolean} [parameters.options.omitSourceMapResources=false] Whether source map resources shall
  * 		be tagged as "OmitFromBuildResult" and no sourceMappingURL shall be added to the minified resource
+ * @param {boolean} [parameters.options.useInputSourceMaps=true] Whether to make use of any existing source
+ * 		maps referenced in the resources to be minified. Use this to preserve the reference to the original
```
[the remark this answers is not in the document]
RandomByte: Done, thanks!

matz3 (commented):
Tests for minify task should be enhanced

RandomByte:
> Tests for minify task should be enhanced

Done

test/lib/tasks/minify.js
```diff
@@ -1,263 +1,164 @@
...
-	t.deepEqual(await resSourceMap.getString(), expectedSourceMap, "Correct source map content");
+	t.is(minifierStub.callCount, 1, "minifier got called once");
+	const minifierCallArgs = minifierStub.firstCall.firstArg;
+	t.deepEqual(minifierCallArgs.resources, ["resource A", "resource B"], "Correct resources provided to processor");
+	t.is(minifierCallArgs.fs, "fs interface", "Correct fs interface provided to processor");
+	t.is(minifierCallArgs.taskUtil, taskUtil, "Correct taskUtil provided to processor");
+	t.deepEqual(minifierCallArgs.options, {
+		addSourceMappingUrl: false,
+		readSourceMappingUrl: false,
+		useWorkers: true
```
matz3, line 122: Maybe add a test case where `useWorkers` is false? (i.e. no taskUtil parameter)

RandomByte: Done 👍

matz3 (approved):

PR #785, opened 2022-08-22
[FEATURE] jsdoc: Improve support for ES6+ syntax 

d3xter666:
JIRA: CPOUI5FOUNDATION-374


test/expected/build/library.j/dest/resources/library/j/dependency-es6-2.js
```diff
@@ -0,0 +1,22 @@
...
+ * ${copyright}
+ */
+
+/**
+ * Covers:
+ * - ArrowFunctionExpression
+ */
+window.someRandomModule ||
+	sap.ui.define(["/.a"], (a) =>
+		a.extend(`aaa`, {
```
matz3, line 11: Should we also add actual JSDoc comments for all those cases so that we can verify whether the api.json contains the expected content? With this change the api.json stays the same.

lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -53,7 +53,14 @@
  */
 
 /* imports */
-const Syntax = require('jsdoc/src/syntax').Syntax;
+const Syntax = {
+	// Those are currently not in the syntax.js file.
+	ChainExpression: "ChainExpression",
+	OptionalMemberExpression: "OptionalMemberExpression",
+	OptionalCallExpression: "OptionalCallExpression",
```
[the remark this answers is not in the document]
d3xter666: This would be the best solution, but didn't we discuss that there might be some prerequisites before that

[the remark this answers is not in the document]
matz3: I'm not completely sure. I guess the version just needs to be aligned beforehand. But maybe @codeworrior can share more insights.

test/expected/build/library.j/dest/resources/library/j/dependency-es6-2.js
d3xter666: Now should be ok

test/expected/build/library.j/dest/resources/library/j/dependency-es6-1.js
```diff
@@ -9,9 +9,24 @@
...
+	 * My super documentation of this class
+	 *
+	 * @extends library.j.Bar
+	 *
+	 * @author SAP SE
+	 * @version ${version}
+	 *
+	 * @public
+	 * @alias library.j.Foo
+	 * @ui5-metamodel text
```
matz3, line 23: Please remove. This tag is about to be removed.

matz3 (changes requested):

test/expected/build/library.j/dest/resources/library/j/dependency-es6-1.js
d3xter666: Thanks! Removed it

lib/processors/jsdoc/lib/ui5/plugin.js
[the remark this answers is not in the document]
matz3: Let's keep it like this for now.

matz3 (approved):
Please use squash and merge and ensure that the message title uses [FEATURE] (as in the PR title)

PR #798, opened 2022-10-12
[FIX] package.json: Downgrade es5-ext dependency

RandomByte:
Downgrade es5-ext dependency as latest versions cause alerts in multiple Antivirus solutions (https://github.com/medikoo/es5-ext/issues/186).

Also downgrade es6-set, since the latest version requires an affected version of es5-ext.

In general, none of these dependencies are required for ui5-builder. We install them transitively through escope. escope has an unreleased (and old) PR for dropping them: https://github.com/estools/escope/pull/124


codeworrior (approved):
LGTM

PR #805, opened 2022-10-26
[INTERNAL] Adjust module path for subpath tasks and processors

flovogt:


matz3 (changes requested):
There are also JSDoc references (e.g. @param in generateJsdoc) that need to be updated

flovogt:
> generateJsdoc

True... Quite hard to catch them. Awesome, @matz3! I've updated the findings for all subpath `processors` and `tasks`.

matz3 (approved):

PR #807, opened 2022-10-28
[INTERNAL] Upgrade supported ECMAScript version to es2022

flovogt:


matz3:
Are we ready to use ES2022? nyc/istanbul also needs to support it in order to parse and instrument our code.

flovogt:
> Are we ready to use ES2022? nyc/istanbul also needs to support it in order to parse and instrument our code.
I've performed a "npm run coverage" run and it looks fine.
Screenshot 2022-10-28 at 11 11 31

matz3 (approved):

PR #811, opened 2022-11-03
[INTERNAL] Fix tests for invalid JSON and themeBuilder

RandomByte:
Node 19 has a new error message for invalid JSON. Instead of
"Unexpected token o in JSON at position 1" it now reads
"Unexpected token 'o', "no json" is not valid JSON"

Theme builder failed due to missing `await` which now showed up due to https://github.com/SAP/ui5-fs/pull/407

### DIFF
--- a/test/lib/processors/themeBuilder.js
+++ b/test/lib/processors/themeBuilder.js
@@ -5,7 +5,7 @@ import fsInterface from "@ui5/fs/fsInterface";
 import themeBuilderProcessor from "../../../lib/processors/themeBuilder.js";
 import {ThemeBuilder} from "../../../lib/processors/themeBuilder.js";
 
-function prepareResources({library} = {}) {
+async function prepareResources({library} = {}) {
 	const input =
 `@someColor: black;
 .someClass {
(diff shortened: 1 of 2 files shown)


test/lib/processors/themeBuilder.js
```diff
@@ -29,7 +29,7 @@ function prepareResources({library} = {}) {
 		string: input
 	});
 
-	memoryAdapter.write(resource);
+	await memoryAdapter.write(resource);
```
matz3, line 32: I guess this problem now became visible after https://github.com/SAP/ui5-fs/pull/407 which adds another "Promise" (although being sync most of the time).

This will probably also reveal issues in other code such as custom tasks, which is good, as this can always lead to issues.
But I'm also thinking about the overhead and impact that the change brings as it creates another promise instance on every resource write.

matz3 (approved):

test/lib/processors/themeBuilder.js
RandomByte: > But I'm also thinking about the overhead and impact that the change brings as it creates another promise instance on every resource write.

Yes, that was also my concern

PR #834, opened 2022-11-07
[INTERNAL] Use Object.create(null) for dynamic maps

matz3:


RandomByte (approved):

flovogt:
For further information about `Object.create(null)` vs `{}` https://davidwalsh.name/object-create-null might be worth to read.

PR #838, opened 2022-11-09
[BREAKING] Removal of manifestBundler and generateManifestBundle

flovogt:
JIRA: CPOUI5FOUNDATION-529


matz3 (changes requested):
The task also should be removed from taskRepository and handled like the other removed tasks (e.g. "createDebugFiles")

flovogt:
> The task also should be removed from taskRepository and handled like the other removed tasks (e.g. "createDebugFiles")

Done

matz3 (dismissed):

matz3 (approved):

PR #851, opened 2022-11-30
[INTERNAL] Adopt consistency check of EnrichedVisitorKeys

matz3:
This change solves the issue of having runtime exceptions when new
VisitorKeys are available via minor version updates of espree.

Instead, a unit test is added to ensure all new VisitorKeys are handled
when updating the ecmaVersion.

We only expect to have new Syntax / VisitorKeys once we update the
ecmaVersion in use, so it shouldn't be an issue when we don't
immediately handle them.


RandomByte (commented):
LGTM

flovogt (approved):
LGTM

codeworrior (approved):

PR #857, opened 2022-12-15
[INTERNAL] Tasks: Handle empty dependencies reader

RandomByte:
If a project has no dependencies, https://github.com/SAP/ui5-fs/pull/455
would lead to tasks being provided with no dependencies reader at all
(opposing to one that does not find any resources, as it is today).


RandomByte:
This PR is mainly to illustrate the impact https://github.com/SAP/ui5-fs/pull/455 would have on (custom) tasks

RandomByte:
Closing since https://github.com/SAP/ui5-fs/pull/455 got closed

PR #858, opened 2022-12-19
Align lib/processors/jsdoc with openui5/lib/jsdoc

codeworrior:
- publish.js: do not write 'module' info for borrowed APIs (part 2)
   This is a follow-up to SAP/openui5@9d3fccd91) which only handled the
   'module' and 'export' properties and only for borrowed methods.
   The same logic now applies to the 'resource' property and also to
   borrowed properties (fields) and events.
   (Cherry picked from SAP/openui5@d6215d715)

- publish.js: write type of @typedefs to api.json
   ...as "extends" - when the @typedef has properties and is not a type alias
   or function prototype
   (Cherry picked from SAP/openui5@5c496b23d)

- publish.js: type parser: accept underscores and $ signs
   ...in type names. Needed in particular for the newly to-be-introduced
   sap.ui.base.Object.$ObjectMetadata etc.
   Related to https://github.com/SAP/ui5-typescript/issues/338
   (Cherry picked from SAP/openui5@7deccb5c7)

- publish.js: do not show a module for namespaces without exports
   When a namespace is not exported by a module, don't show a module
   for the namespace in the SDK. This prevents useless module information
   for purely virtual namespaces like sap.ui.base, sap.ui.model.json etc.
   (Cherry picked from SAP/openui5@396262641)

### DIFF
--- a/lib/processors/jsdoc/lib/ui5/template/publish.cjs
+++ b/lib/processors/jsdoc/lib/ui5/template/publish.cjs
@@ -3067,6 +3067,12 @@ function createAPIJSON4Symbol(symbol, omitDefaults) {
 		if ( symbol.params || symbol.returns || symbol.exceptions ) {
 			methodSignature(symbol);
 		} else if ( symbol.properties && symbol.properties.length > 0 ) {
+			if ( symbol.type ) { // "type" of a typedef defines its inheritance base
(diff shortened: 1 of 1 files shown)


flovogt (commented):
LGTM

RandomByte (approved):

PR #870, opened 2023-01-17
[BREAKING] Drop support for node 16

flovogt:
This PR drops support for Node 16. Only node versions higher than `18.0.0` are supported.


flovogt:
Decided to support Node.js version 16.18.0 and above

PR #874, opened 2023-01-20
[INTERNAL] stringReplacer: Replace using string instead of stream

RandomByte:
This allows the processor to only modify a resource if necessary (as in,
a replacement has been applied). Currently, in a standard library build,
this processor is for example the only one to modify JSON files.
However, they often do not contain any of the searched placeholders.

https://github.com/SAP/ui5-fs/pull/472 will enable optimizations for
unmodified resources. So we strive for reducing unnecessary resource
modification across the whole tooling.

### DIFF
--- a/test/lib/processors/stringReplacer.js
+++ b/test/lib/processors/stringReplacer.js
@@ -1,39 +1,23 @@
 import test from "ava";
-import {Readable} from "node:stream";
+import _sinon from "sinon";
 import stringReplacer from "../../../lib/processors/stringReplacer.js";
 
-const getStringFromStream = (stream) => {
-	return new Promise((resolve, reject) => {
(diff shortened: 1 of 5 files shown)


test/lib/processors/stringReplacer.js
```diff
@@ -1,39 +1,46 @@
 import test from "ava";
-import {Readable} from "node:stream";
+import _sinon from "sinon";
```
flovogt, line 2: is it necessary to use here "_" ? The const declarations in the the tests are created in a new scope isn't?

flovogt: Ack.

flovogt (approved):

PR #883, opened 2023-02-07
[INTERNAL] lib/processors/jsdoc: enhance list of built-in types and remove direct usage of taffydb

flovogt:
TODO: most parts of the hard coded lists should be replaced with content
from the npm package 'globals'.

Cherry picked from SAP/openui5@b53bbcc84.

----------------------------

Avoid the use of TaffyDB APIs in the active publishing variants
(apijson, apixml, fullapixml) as not all TaffyDB APIs are supported by
the replacement 'Salty' in JSDoc 4.0.0.

The get() API is still used, all other operations are replaced with
native array operations.

Legacy publishing variants (apijs, full, public, demokit,
demokit-internal) are not adapted as they are no longer functional.
They'll be cleaned up in a later change.

JIRA: CPOUI5FOUNDATION-595

Cherry picked from SAP/openui5@187bc3e07.


RandomByte (approved):
Personally, I'd prefer to have one commit per cherry-pick to make it easier to compare to the commits in OpenUI5. The title of commit [[INTERNAL] JSDoc: prepare for the use of JSDoc 4.0.0](https://github.com/SAP/openui5/commit/187bc3e0708aa9d3be360343c4af182dc992d4ea) now doesn't appear anywhere here.

I think that's also what Frank usually does: https://github.com/SAP/ui5-builder/pull/878

Besides that, looks good and contains all the missing change 👍

PR #901, opened 2023-05-10
[FIX] lbt/Resolver: ensure stable module order in generated bundleInfo

codeworrior:
When a section of type 'bundleInfo' is generated as part of a bundle, the order of the modules in each bundle info is not relevant, but should be stable across different runs.

So far, the order of modules was dependent on the module iteration order of the resource pool and could be random due to I/O timing.

### DIFF
--- a/lib/lbt/bundle/Resolver.js
+++ b/lib/lbt/bundle/Resolver.js
@@ -295,7 +295,9 @@ class BundleResolver {
 								return modules;
 							});
 						}
-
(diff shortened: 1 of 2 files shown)


codeworrior:
I decided against my own proposal to sort the modules in the Builder's writeBundleInfo only, as the additional bundle info object returned by the Builder was not stable, too. Sorting the modules in the Resolver instead, made the order stable in both places.

d3xter666 (approved):

matz3 (approved):

PR #907, opened 2023-06-07
[INTERNAL] Support imported Core dependency when analyzing library.js

d3xter666:
JIRA: CPOUI5FOUNDATION-476

Analyzes library.js files for dependencies that might be able to initialize a UI5 library. Currently possible options now are:
- `sap.ui.getCore().initLibrary()`- already available
- 
  ```js 
  sap.ui.define(['sap/ui/core/Core',], function(Core) {
      var thisLib = Core.initLibrary({
      ...
  ```
- 
  ```js 
  sap.ui.define(['sap/ui/core/Lib',], function(Library) {
      var thisLib = Library.init({
      ...
  ```

### DIFF
--- a/lib/lbt/analyzer/analyzeLibraryJS.js
+++ b/lib/lbt/analyzer/analyzeLibraryJS.js
@@ -25,8 +25,8 @@ async function analyze(resource) {
 		interfaces: []
 	};
 
-	function visit(node) {
-		if (isInitLibraryCall(node, resource)) {
+	function visit(node, sapUiDefineCall) {
+		if (isInitLibraryCall(node, sapUiDefineCall)) {
 			node.arguments[0].properties.forEach( (prop) => {
 				if (prop.type === Syntax.SpreadElement) {
 					// SpreadElements are currently not supported
@@ -64,11 +64,11 @@ async function analyze(resource) {
 		for ( const key of VisitorKeys[node.type] ) {
 			const child = node[key];
 			if ( Array.isArray(child) ) {
-				if ( child.some(visit) ) {
+				if ( child.some((c) => visit(c, sapUiDefineCall)) ) {
 					return true;
 				}
 			} else if ( child ) {
-				if ( visit(child) ) {
+				if ( visit(child, sapUiDefineCall) ) {
 					return true;
 				}
 			}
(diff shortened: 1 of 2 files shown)


lib/lbt/analyzer/analyzeLibraryJS.js
```diff
@@ -5,6 +5,11 @@ import {getLogger} from "@ui5/logger";
 const log = getLogger("lbt:analyzer:LibraryJS");
 
 const CALL__SAP_UI_GETCORE = ["sap", "ui", "getCore"];
+const AMD__LIB_INIT_RESOURCES = {
+	// Depenedency definition: inti library method name
```
matz3, line 9:
```suggestion
	// Depenedency definition: init library method name
```

lib/lbt/analyzer/analyzeLibraryJS.js
```diff
@@ -82,6 +82,60 @@ async function analyze(resource) {
 	return libInfo;
 }
 
+const initLibraryDefintions = [];
```
matz3, line 85: This module should not hold state in between calls.
I see potential side-effects when this module is called multiple times as this array is not cleared.

lib/lbt/analyzer/analyzeLibraryJS.js
```diff
@@ -82,6 +82,60 @@ async function analyze(resource) {
...
+
+		for (let i = 0; i < amdFnArguments.length; i++) {
+			const fnArgName = amdFnArguments[i].name; // Name of the variable
+			const amdDef = amdDepDefinitions[i].value;
+			if (AMD__LIB_INIT_RESOURCES[amdDef]) {
+				// Add just the ones that are available
+				initLibraryDefintions.push([fnArgName, AMD__LIB_INIT_RESOURCES[amdDef]]);
+			}
+		}
+	}
```
matz3, line 118: I expect that there's a better solution than having a side-effect within this `isInitLibraryCall` function.
Also, wouldn't it make sense to use `lbt/calls/SapUiDefine.js` to re-use the analysis of the sap.ui.define call?

matz3 (changes requested):

lib/lbt/analyzer/analyzeLibraryJS.js
d3xter666: Thanks. Switched to sapUiDefine analyzer

lib/lbt/analyzer/analyzeLibraryJS.js
matz3: I wasn't referring to the two-step approach which is indeed needed. I'm rather questioning why the variable is declared in the global scope of the module rather than within the "analyze" function scope.

Furthermore I don't think it's a good practice to collect that information in the same function that is supposed to tell whether a node is an init call. This should be separated.

test/lib/lbt/analyzer/analyzeLibraryJS.js
```diff
@@ -154,3 +157,96 @@ sap.ui.define([
...
+	return thisLib;
+});`;
+
+	mockResource = createMockResource(libraryJS, librayJSPath, librayName);
+	result = await analyzeLibraryJSWithStubbedLogger(mockResource);
+	t.is(errorLogStub.callCount, 0, "Error log is not called");
+	t.is(result.elements[0], "library.test.MenuItem", "The libraryjs is correctly analyzed");
+});
+
+test.serial("analyze: library.js with unknonwn initLibrary call", async (t) => {
```
matz3, line 217:
```suggestion
test.serial("analyze: library.js with unknown initLibrary call", async (t) => {
```

lib/lbt/analyzer/analyzeLibraryJS.js
```diff
@@ -76,8 +82,76 @@ async function analyze(resource) {
...
+			const child = node[key];
+			if (Array.isArray(child)) {
+				return child
+					.map(analyzeSapUiDefineCalls)
+					.filter((potentialCall) => potentialCall && potentialCall.dependencies)[0];
+			} else if (child) {
+				return analyzeSapUiDefineCalls(child);
+			}
+		}
+	}
```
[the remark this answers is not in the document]
d3xter666: Thanks!

codeworrior:
@d3xter666 I'm not able to view the diff, 182868ec82d3cf31b2be7f729711b8dae75f23e7d65faf4b88c93e151362d144 seems not to exist in this repo (maybe only before the last force-push?)

d3xter666:
Hi @codeworrior,

It's actually the last commit where I switched to the common `define` analysis.

Please, take a look at the `analyzeSapUiDefineCalls` method in [6d51dd0ac3ed24e37b954a765e34097078200213](https://github.com/SAP/ui5-builder/pull/907/commits/6d51dd0ac3ed24e37b954a765e34097078200213).
The former is the general one and the current refactoring is the alignment.

Best regards

lib/lbt/analyzer/analyzeLibraryJS.js
```diff
@@ -19,13 +27,17 @@ async function analyze(resource) {
...
+	visit( rootNode );
+
 	function visit(node) {
-		if ( node.type == Syntax.CallExpression &&
-				node.callee.type === Syntax.MemberExpression &&
-				isMethodCall(node.callee.object, CALL__SAP_UI_GETCORE) &&
-				isIdentifier(node.callee.property, "initLibrary") &&
-				node.arguments.length === 1 &&
-				node.arguments[0].type === Syntax.ObjectExpression ) {
+		// Define call would always be at the top of th tree, so it'd
```
codeworrior, line 36: small typo: th(e)

codeworrior (commented):
Two small typos and two potential optimisations. The 'if' I would recommend to add, the second optimisation is just an idea and shouldn't be necessary, taking the low number of executions of this analyser into account.

Reg. the overall structure, I'm in favour of the current one. There's no rule that `sap.ui.define` MUST be the first top-level statement. It's a best practice for AMD/UI5, but not mandatory. The use case that @matz3 presented (`sap.ui.loader.config` call in front of the `sap.ui.define` call) is a good and valid example.

lib/lbt/analyzer/analyzeLibraryJS.js
d3xter666: Thanks

lib/lbt/analyzer/analyzeLibraryJS.js
```diff
@@ -1,10 +1,18 @@
...
 import {getPropertyKey, isMethodCall, isIdentifier, getStringArray} from "../utils/ASTUtils.js";
+import SapUiDefineCall from "../calls/SapUiDefine.js";
 import {getLogger} from "@ui5/logger";
 const log = getLogger("lbt:analyzer:LibraryJS");
 
 const CALL__SAP_UI_GETCORE = ["sap", "ui", "getCore"];
+const CALL_SAP_UI_DEFINE = ["sap", "ui", "define"];
+const CALL_AMD_DEFINE = ["define"];
+const AMD__LIB_INIT_RESOURCES = [
+	// Depenedency definition, init library method name
```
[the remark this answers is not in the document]
d3xter666: Thanks!

lib/lbt/analyzer/analyzeLibraryJS.js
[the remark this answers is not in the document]
d3xter666: Thanks!

lib/lbt/analyzer/analyzeLibraryJS.js
```diff
@@ -19,13 +27,17 @@ async function analyze(resource) {
...
-				isMethodCall(node.callee.object, CALL__SAP_UI_GETCORE) &&
-				isIdentifier(node.callee.property, "initLibrary") &&
-				node.arguments.length === 1 &&
-				node.arguments[0].type === Syntax.ObjectExpression ) {
+		// Define call would always be at the top of the tree, so it'd
+		// be the first significant thing that's been found.
+		sapUiDefineCall = sapUiDefineCall || analyzeSapUiDefineCalls(node);
+
+		if (isInitLibraryCall(node)) {
 			node.arguments[0].properties.forEach( (prop) => {
```
matz3, line 41: I can't put a comment below, but the `log.error` could be improved to properly reflect the actual call or maybe mention something more general than `sap.ui.getCore().initLibrary`.

lib/lbt/analyzer/analyzeLibraryJS.js
```diff
@@ -19,13 +27,17 @@ async function analyze(resource) {
...
 	function visit(node) {
-		if ( node.type == Syntax.CallExpression &&
-				node.callee.type === Syntax.MemberExpression &&
-				isMethodCall(node.callee.object, CALL__SAP_UI_GETCORE) &&
-				isIdentifier(node.callee.property, "initLibrary") &&
-				node.arguments.length === 1 &&
-				node.arguments[0].type === Syntax.ObjectExpression ) {
+		// Define call would always be at the top of th tree, so it'd
+		// be the first significant thing that's been found.
+		sapUiDefineCall = sapUiDefineCall || analyzeSapUiDefineCalls(node);
```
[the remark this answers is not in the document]
d3xter666: Done

lib/lbt/analyzer/analyzeLibraryJS.js
d3xter666: Thanks!

matz3 (commented):
LGTM

codeworrior (approved):
...look good to me, too

PR #914, opened 2023-07-06
[INTERNAL] Optimise themeBuild with workerpool

d3xter666:
JIRA: CPOUI5FOUNDATION-614


lib/tasks/buildThemes.js
```diff
@@ -128,19 +166,45 @@ export default async function({
...
+					string: await res.getString(),
+					name: res.getName(),
+					path: res.getPath(),
+				};
+			})
+		);
+
+	const processedResources = await buildTheme({
+		allResources: await makeTransferableResource(allResources),
+		themeResources: await makeTransferableResource(themeResources),
```
matz3, line 198: I expect it to be very expensive to always transfer all files to the worker. What about just requesting the needed files from within the worker instead? I'd like to see how that would compare performance-wise.

lib/tasks/buildThemes.js
```diff
@@ -66,7 +66,7 @@ async function buildThemeInWorker(options, taskUtil) {
 export default async function({
 	workspace, dependencies, taskUtil,
 	options: {
-		projectName, inputPattern, librariesPattern, themesPattern, compress, cssVariables, useWorkers = true
+		projectName, inputPattern, librariesPattern, themesPattern, compress, cssVariables, useWorkers = false
```
flovogt, line 69: add a `TODO for V4` to change default to true

d3xter666: Done

lib/tasks/buildThemes.js
```diff
@@ -128,19 +164,51 @@ export default async function({
...
+			throw new Error(`buildThemes: Option 'useWorkers' requires a taskUtil instance to be provided`);
 		}
-	});
+		buildTheme = buildThemeInWorker;
+	} else {
+		// Do not use workerpool
+		buildTheme = (await import("../processors/themeBuilderWorker.js")).default;
+	}
+
+	// All the relevant resources need to be made available for the less parser within the workerpool
```
matz3, line 182: All those things are not relevant when `useWorkers=false`. I'd expect that the new handling is only applied when using workers. Otherwise it might cause unnecessary overhead for existing use cases.

lib/tasks/buildThemes.js
```diff
@@ -128,19 +164,51 @@ export default async function({
...
+	const allResources = await combo.byGlob("/resources/**/{*.less,.theming,img/**,img-RTL/**}");
+	const makeTransferableResource = async (resourceCollection) =>
+		Promise.all(
+			resourceCollection.map(async (res) => ({
+				transferable: await res.getBuffer(), // Buffer -> Uint8Array (transferable object)
+				path: res.getPath()
+			}))
+		);
+	let allResourcesTransferable;
+	if (allResources) {
```
matz3, line 192: How can `allResources` become falsy? `byGlob` should always return an Array, even when it's empty.

lib/processors/themeBuilderWorker.js
```diff
@@ -0,0 +1,88 @@
...
+}) {
+	const fsThemeResources = themeResources.map((res) => {
+		// res.transferable is an Uint8Array object and needs to be casted
+		// to a Buffer in order to be read correctly.
+		return createResource({path: res.path, buffer: Buffer.from(res.transferable)});
+	});
+	const fsAllResources = allResources.map((res) => {
+		// res.transferable is an Uint8Array object and needs to be casted
+		// to a Buffer in order to be read correctly.
+		return createResource({path: res.path, buffer: Buffer.from(res.transferable)});
```
matz3, line 34: Duplicate code which can easily be avoided

lib/processors/themeBuilderWorker.js
```diff
@@ -0,0 +1,88 @@
...
+	});
+
+	return makeTransferableResource(result);
+}
+
+/**
+ * Cast @ui5/fs/Resource into an Uint8Array transferable object
+ *
+ * @private
+ * @param {@ui5/fs/Resource} resourceCollection
```
matz3, line 58: Incorrect type

lib/processors/themeBuilderWorker.js
```diff
@@ -0,0 +1,88 @@
...
+ * @function default
+ * @static
+ *
+ * @param {object} parameters Parameters
+ * @param {object} parameters.allResources Array of Uint8Array transferable objects
+ * 	with all the resources needed to build the css files. By nature those are @ui5/fs/Resource.
+ * @param {object} parameters.themeResources Input array of Uint8Array transferable objects
+ * 	that are the less sources to build upon. By nature those are @ui5/fs/Resource.
+ * @param {object} parameters.options Less compiler options
+ * @returns {Promise<Uint8Array[]>}
```
matz3, line 19: Type differs from type described for `makeTransferableResource`.

lib/processors/themeBuilderWorker.js
```diff
@@ -0,0 +1,88 @@
...
+async function makeTransferableResource(resourceCollection) {
+	return Promise.all(
+		resourceCollection.map(async (res) => ({
+			transferable: await res.getBuffer(),
+			path: res.getPath()
+		}))
+	);
+}
+
+async function addResourcesToCollection(resources) {
```
matz3, line 70: Function name does not reflect what it does. It creates a new collection but the name suggest that the resources are added to a collection that can be passed (at least that's my personal understanding).

lib/processors/themeBuilderWorker.js
```diff
@@ -0,0 +1,88 @@
...
+	});
+	const fsAllResources = allResources.map((res) => {
+		// res.transferable is an Uint8Array object and needs to be casted
+		// to a Buffer in order to be read correctly.
+		return createResource({path: res.path, buffer: Buffer.from(res.transferable)});
+	});
+
+	const fsReader = await addResourcesToCollection(fsAllResources);
+	const wCollection = createWriterCollection({
+		name: `theme builder worker- build resources collection`,
```
matz3, line 39:
```suggestion
		name: `theme builder worker - build resources collection`,
```

lib/processors/themeBuilderWorker.js
```diff
@@ -0,0 +1,88 @@
...
+	return Promise.all(
+		resourceCollection.map(async (res) => ({
+			transferable: await res.getBuffer(),
+			path: res.getPath()
+		}))
+	);
+}
+
+async function addResourcesToCollection(resources) {
+	const fsTarget = createAdapter({
```
matz3, line 71: Why `fsTarget`? It's not a FileSystem adapter.

matz3 (changes requested):
Tests missing for new functionality

lib/tasks/buildThemes.js
```diff
@@ -67,7 +67,7 @@ export default async function({
 	workspace, dependencies, taskUtil,
 	options: {
 		projectName, inputPattern, librariesPattern, themesPattern, compress,
-		cssVariables, useWorkers = false /* TODO: for UI5Tooling v4 switch to "true" */
+		cssVariables, useWorkers = false /* TODO: for UI5 Tooling v4 switch to "true" */
```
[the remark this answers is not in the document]
d3xter666: Done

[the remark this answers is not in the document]
flovogt: can we not enable it per default?

lib/processors/themeBuilderWorker.js
d3xter666: Thanks!

lib/processors/themeBuilderWorker.js
d3xter666: Indeed. It used to be as in the beginning we were storing the resources on the FS, but now it's a MemoryAdapter.
Thanks!

lib/processors/themeBuilderWorker.js
d3xter666: Done

lib/processors/themeBuilderWorker.js
d3xter666: Thanks

lib/tasks/buildThemes.js
d3xter666: I guess, it's because of the tests. I've been missled by the stubs. Thanks!

lib/tasks/buildThemes.js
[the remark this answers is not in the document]
d3xter666: I saw it this way for the [minifier](https://github.com/SAP/ui5-builder/blob/main/lib/processors/minifier.js#L75)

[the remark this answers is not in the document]
matz3: We can only enable it via the build definitions

lib/tasks/buildThemes.js
d3xter666: Done

lib/processors/themeBuilderWorker.js
d3xter666: Done

lib/processors/themeBuilderWorker.js
```diff
@@ -0,0 +1,94 @@
...
+	return makeTransferableResource(result);
+}
+
+/**
+ * Casts @ui5/fs/Resource-s into an Uint8Array transferable object
+ *
+ * @param {@ui5/fs/Resource[]} resourceCollection
+ * @returns {Promise<object[]>}
+ */
+export async function makeTransferableResource(resourceCollection) {
```
matz3, line 53: Isn't this the reversed step of `uint8ToResources`?
If so, why not calling it `resourcesToUint8`?

d3xter666: Done

PR #922, opened 2023-07-13
[FIX] minifier: Improve performance

matz3:
Using Uint8Array (Transferable object) to reduce overhead of
transporting large strings between threads.

### DIFF
--- a/lib/processors/minifierWorker.js
+++ b/lib/processors/minifierWorker.js
@@ -40,9 +40,9 @@ export default async function execMinification({
 	sourceMapOptions
 }) {
 	try {
-		const result = await minify({
+		const minifyResult = await minify({
 			// Use debug-name since this will be referenced in the source map "sources"
-			[dbgFilename]: Buffer.from(code).toString()
+			[dbgFilename]: Buffer.from(code.buffer, code.byteOffset, code.length).toString()
 		}, {
 			output: {
 				comments: copyrightCommentsAndBundleCommentPattern,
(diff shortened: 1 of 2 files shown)


matz3:
Note: It's not clear whether this is really improving performance. Benchmarks need to be performed.

lib/processors/minifierWorker.js
```diff
@@ -40,9 +40,9 @@ export default async function execMinification({
 	sourceMapOptions
 }) {
 	try {
-		return await minify({
+		const result = await minify({
 			// Use debug-name since this will be referenced in the source map "sources"
-			[dbgFilename]: code
+			[dbgFilename]: Buffer.from(code).toString()
```
d3xter666, line 45: I got some wird results when calling `Buffer.toString()` without encoding in https://github.com/SAP/ui5-builder/pull/914 when (de-)serializing `@ui5/fs/Resources`.
Maybe, it's a good point to add an encoding. I guess, if those are just `js` files, we can asume `utf8`

matz3: That's strange as we're using it like that ever since in ui5-fs: https://github.com/SAP/ui5-fs/blob/f378330dc1da077b6f945d59c41592b776a7001e/lib/Resource.js#L175

matz3: With binary files such as images, it doesn't work for obvious reasons, though.

d3xter666: Sorry, I was refering to this one: https://github.com/SAP/ui5-fs/blob/f378330dc1da077b6f945d59c41592b776a7001e/lib/Resource.js#L185

matz3: If the first argument is a string, the encoding defaults to utf-8. At least that's how the function is documented in Node.js

d3xter666 (approved):

matz3:
As stated above, this still needs further testing on whether and how it improves performance.

I forgot to mark this as WIP, and it shouldn't be merged, yet.

matz3:
It turned out that this approach doesn't work as expected and it's fastest to just transfer the code as string.

Also see https://nodejs.org/api/worker_threads.html#considerations-when-transferring-typedarrays-and-buffers
> For Buffer instances, specifically, whether the underlying ArrayBuffer can be transferred or cloned depends entirely on how instances were created, which often cannot be reliably determined.

PR #924, opened 2023-07-14
[FIX] Bundling: Detect manifest.json dependency of Components / Libraries

matz3:


lib/lbt/analyzer/ComponentAnalyzer.js
```diff
@@ -54,6 +54,7 @@ class ComponentAnalyzer {
 		try {
 			const manifestResource = await this._pool.findResource(manifestName).catch(() => null);
 			if ( manifestResource ) {
+				info.addDependency(manifestName);
```
codeworrior, line 57: I think we need to discuss whether the existence of a manifest.json should be the indicator or the manifest: json entry in the metadata or both. 

* If a manifest.json exists, manifest-first would work even if the component does not have the metadata entry
* if the metadata entry exists, a manifest.json would be requested even if it doesn't exist

matz3: I agree. My idea was to always add the dependency because currently the `manifest.json` is always taken into account for dependency analysis.

I suggest to merge this change as-is and to follow-up on the ComponentAnalyzer as a whole, which also seems to lack detection of new features (e.g. in routing).

matz3: I've created https://github.com/SAP/ui5-builder/pull/954 which only covers libraries.

PR #925, opened 2023-07-14
[INTERNAL] Optimise themeBuild with workerpool via fsinterface

d3xter666:
JIRA: CPOUI5FOUNDATION-614

This change enhances https://github.com/SAP/ui5-builder/pull/914 with a fsInterface like objects that "read" the information through `postMessages` via `MessageChannel -> MessagePort`


lib/processors/themeBuilderWorker.js
```diff
@@ -92,3 +72,129 @@ if (!workerpool.isMainThread) {
 		execThemeBuild
 	});
 }
+
+
+export class FsMainThreadInterface {
+	#comPort = null;
+	#fsInterfaceReader = null;
+	#cache = Object.create(null);
```
RandomByte, line 80: Maybe I'm missing something but isn't the relationship from FsMainThreadInterface to FsWorkerThreadInterface 1:1?

So whatever is requested in the worker will also be cached there (which I think makes sense, but this cache in the main thread will never get any hits?

It might make sense to have an `@ui5/fs` level cache for the combo reader that is shared by all FaMainThreadInterface instances, since many worker will probably access the same resources and they are not expected to change while the task is running.

d3xter666: Indeed, that was my point initially. 
I'll try to make the `FaMainThreadInterface` a singleton that manages all the connections, but keeps a cache for the already resolved resources, because that's the biggest possible improvement here. Each `Resource` gets resolved by the combo reader's globbing for up to 40ms. Caching in that case for big builds like `themebuild_sap_belize` might squize some more milliseconds from the build step.

lib/processors/themeBuilderWorker.js
```diff
@@ -91,3 +72,145 @@ if (!workerpool.isMainThread) {
...
+	#onClose() {
+		this.#comPort.close();
+		this.#cache = null;
+	}
+
+	#doRequest({what, fsPath, options}, callback) {
+		const cacheKey = [fsPath, what].join("-");
+
+		if (this.#cache[cacheKey]) {
+			callback(null, this.#cache[cacheKey]);
```
matz3, line 194: What about the error? I guess it is intended to not cache fs calls with an error, right? This totally makes sense to me but unfortunately this is not easily visible in the code. Some comments would help to understand why the caching behaves this way.

lib/processors/themeBuilderWorker.js
```diff
@@ -91,3 +72,145 @@ if (!workerpool.isMainThread) {
...
+			throw new Error("Communication port is mandatory argument");
+		}
+
+		this.#comPort = comPort;
+		this.#comPort.on("message", this.#onMessage.bind(this));
+		this.#comPort.on("close", this.#onClose.bind(this));
+	}
+
+	#onMessage(e) {
+		const cbObject = this.#callbacks.find((cb) => cb.what === e.what && cb.fsPath === e.fsPath);
```
matz3, line 174: At first reading I was not able to guess the meaning of `what` in here. Further down in the code it becomes clear, that it is the fs function name to be invoked.

lib/tasks/buildThemes.js
```diff
@@ -33,8 +33,10 @@ function getPool(taskUtil) {
 	return pool;
 }
 
-async function buildThemeInWorker(options, taskUtil) {
-	return getPool(taskUtil).exec("execThemeBuild", [options]);
+async function buildThemeInWorker(taskUtil, options, transferList) {
+	const toTransfer = transferList ? {transfer: transferList} : undefined;
+
+	return getPool(taskUtil).exec("execThemeBuild", [options], toTransfer);
```
matz3, line 39: I wonder why the last argument hasn't been used in https://github.com/SAP/ui5-builder/pull/914. Does that mean that it never really used transferable objects in a correct way? I also just looked up https://github.com/josdejong/workerpool/blob/master/examples/transferableObjects.js / https://github.com/josdejong/workerpool/blob/master/examples/workers/transferableObjectsWorker.js which is different to the current approach of transferring objects.

matz3: AFAICS PR #914 is using the "copy" approach, as defined in https://github.com/josdejong/workerpool/blob/8219f7deaf535310cb8c3f273cda8e047959502c/examples/transferableObjects.js#L24-L29.

lib/processors/themeBuilderWorker.js
d3xter666: I've renamed it to "action". Maybe, makes more sense

lib/tasks/buildThemes.js
d3xter666: As we discussed, transferring objects wouldn't add benefits in terms of performance, but would make the code more complex. Leaving the copies for this implementation

lib/processors/themeBuilderWorker.js
d3xter666: As we have discussed, it makes sense here also to cache the errors.

d3xter666:
Current build fails, because of the `taskUtil` that is provided in most of the cases, so the workerpool execution becomes there the new "default". I'll adjust those tests later

lib/processors/themeBuilderWorker.js
```diff
@@ -0,0 +1,291 @@
...
+	return resourcesToUint8(result);
+}
+
+/**
+ * Casts @ui5/fs/Resource-s into an Uint8Array transferable object
+ *
+ * @param {@ui5/fs/Resource[]} resourceCollection
+ * @returns {Promise<object[]>}
+ */
+export async function resourcesToUint8(resourceCollection) {
```
RandomByte, line 43: Since we use plain buffers now instead of Uint8Arrays, maybe we should rename these functions?

I'm thinking they might even become standard functions on the Resource class if we need this more often. The alternative is of course to change the themeBuilder implementation to not use Resources but plain objects.

```suggestion
export async function serializeResources(resourceCollection) {
```

lib/processors/themeBuilderWorker.js
```diff
@@ -0,0 +1,291 @@
...
+ * Casts Uint8Array into @ui5/fs/Resource-s transferable object
+ *
+ * @param {Promise<object[]>} resources
+ * @returns {@ui5/fs/Resource[]}
+ */
+export function uint8ToResources(resources) {
+	return resources.map((res) => {
+		// res.transferable is an Uint8Array object and needs to be cast
+		// to a Buffer in order to be read correctly.
+		return createResource({path: res.path, buffer: Buffer.from(res.transferable)});
```
RandomByte, line 62: `transferable` is already a buffer now

```suggestion
		return createResource({path: res.path, buffer: res.transferable));
```

RandomByte (changes requested):
Looking pretty good!

lib/processors/themeBuilderWorker.js
```diff
@@ -0,0 +1,291 @@
...
+
+	stat(fsPath, callback) {
+		this.#doRequest({action: "stat", fsPath}, callback);
+	}
+
+	readdir(fsPath, callback) {
+		this.#doRequest({action: "readdir", fsPath}, callback);
+	}
+
+	mkdir(fsPath, callback) {
```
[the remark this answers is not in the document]
d3xter666: Nope, just the readFile and stats is used here. However, I wanted to make it fully compatible with the fsReaderInterface.
If you insist, I would remove the ones that we do not use 😃

lib/processors/themeBuilderWorker.js
d3xter666: Unfortunatelly, this is not correct. The case is that when it goes through the `postMessage` on the other side it is received as `Uint8Array` and that's why we need to explicitly cast it back to `Buffer`

lib/processors/themeBuilderWorker.js
```diff
@@ -0,0 +1,291 @@
...
+			});
+		}
+
+		let fromCache = await this.#cache[cacheKey];
+		// The result from an "action" is usually a simple data type,
+		// except for the stat. In that case we have a Stat object
+		// that has some special methods. In this case, we do not need those,
+		// but just to check whether there's something. The issue is that postMessage
+		// might not be able to serialize the Stat.
+		fromCache = JSON.parse(JSON.stringify(fromCache));
```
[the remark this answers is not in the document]
d3xter666: Done

lib/processors/themeBuilderWorker.js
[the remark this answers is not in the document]
RandomByte: Ok, let's remove it then 😉 

`fsInterface` is already not complete and only provides what we need so far (most prominently missing a `write` function)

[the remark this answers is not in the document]
d3xter666: Done

lib/processors/themeBuilderWorker.js
```diff
@@ -0,0 +1,281 @@
...
+	 * @param {object} parameters.options Options for "readFile" action
+	 */
+	async #doRequest(comPort, {action, fsPath, options}) {
+		const cacheKey = [fsPath, action].join("-");
+		if (!this.#cache[cacheKey]) {
+			this.#cache[cacheKey] = new Promise((res) => {
+				if (action === "readFile") {
+					this.#fsInterfaceReader.readFile(fsPath, options, (error, result) => res({error, result}));
+				} else if (action === "stat") {
+					this.#fsInterfaceReader.stat(fsPath, (error, result) => res({error, result}));
```
RandomByte, line 165: (Future?) Optimization: I guess now we could do the `JSON.parse(JSON.stringify())` once per stat result right here and cache the object.

lib/processors/themeBuilderWorker.js
```diff
@@ -0,0 +1,281 @@
...
+	 * @param {MessagePort} comPort The communication channel
+	 * @param {object} parameters
+	 * @param {string} parameters.action Action to perform. Corresponds to the names of
+	 * 	the public methods of "@ui5/fs/fsInterface" and triggers this method of the
+	 *	"@ui5/fs/fsInterface" instance.
+	 * @param {string} parameters.fsPath Path of the Resource
+	 * @param {object} parameters.options Options for "readFile" action
+	 */
+	async #doRequest(comPort, {action, fsPath, options}) {
+		const cacheKey = [fsPath, action].join("-");
```
RandomByte, line 159: Missed this spot in my first review

```suggestion
		const cacheKey = `${fsPath}-${action}`;
```

RandomByte (approved):
Only minor comments which we could also tackle later, great work 🚀

RandomByte (dismissed):

lib/processors/themeBuilderWorker.js
d3xter666: Ah, nice catch! Thanks!

RandomByte (approved):

PR #927, opened 2023-07-18
Align jsdoc template with openui5

codeworrior:


d3xter666 (approved):
LGTM
I just checked and can confirm that this alignment does not break the generated documentation.

PR #929, opened 2023-07-18
[FIX] generateJsdoc: Also process resources created by preceeding tasks

RandomByte:
End the use of byGlobSource and alwasy use byGlob. None of the standard
tasks executed before it add any resources. So no overhead is to be
expected from this.

On the other hand, the use of custom task extensions (for exmaple for
transpilation) requires the generateJsdoc to process resources the task
has created.

This was the last usage of the byGlobSource API of the DuplexCollection.
Therefore the API shall be deprecated in ui5-fs and removed in the
future.

### DIFF
--- a/lib/tasks/jsdoc/generateJsdoc.js
+++ b/lib/tasks/jsdoc/generateJsdoc.js
@@ -154,12 +154,7 @@ const utils = {
 			virBasePath: "/resources/"
 		});
 
-		let allResources;
-		if (workspace.byGlobSource) { // API only available on duplex collections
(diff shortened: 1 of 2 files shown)


petermuessig (commented):
LGTM

d3xter666 (approved):
LGTM

matz3 (approved):
LGTM. I could not find a specific reason for using that API, and I also couldn't find a case where this causes a regression, even with the initial CLI v1.2.2 that introduced the jsdoc command.

PR #937, opened 2023-09-07
[INTERNAL] dependabot auto-merge: update trigger types and conditions

matz3:
JIRA: CPOUI5FOUNDATION-717


d3xter666 (approved):
I am fine with the change, but it seems we're in the same sitiation with Node v20.6 and maybe we should use Node v20.5 or wait for a NodeJs fix

matz3:
Right. And as there's still no patch available, I've created https://github.com/SAP/ui5-builder/pull/938

PR #943, opened 2023-10-06
[FEATURE] Validate apiVersion property in Library.init

d3xter666:
JIRA: CPOUI5FOUNDATION-730


RandomByte (approved):
LGTM + positive manual test in the core library 👍

PR #945, opened 2023-10-10
[INTERNAL] Bundler: Ignore non-JS resources when creating module name mapping

RandomByte:
This leads to issues where non-JS resources tagged as "IsDebugVariant" (like source map resources) where processed since ModuleName#getNonDebugName can only handle .js and .css extensions.


flovogt (approved):

flovogt (changes requested):
As far as I see, you have started a smoke test run (id: 1518), so smoke tests fails. lets have a deeper look into it together.

matz3 (approved):

RandomByte:
Smoke tests will still fail after this PR is merged. However instead of failing due to an exception (which was caused by this not being able to resolve non-debug names for .map resources), it's only failing because of new warning logs now.

PR #951, opened 2023-11-01
[FEATURE] depCache bundling mode

d3xter666:
JIRA: CPOUI5FOUNDATION-744

Dependency cache information that lists modules and their dependencies of all types: JS, declarative views/fragments.
Only the dependencies of the modules are stored as 'depCache' configuration.

### DIFF
--- a/lib/lbt/bundle/AutoSplitter.js
+++ b/lib/lbt/bundle/AutoSplitter.js
@@ -203,7 +203,6 @@ class AutoSplitter {
 				break;
 			case SectionType.DepCache:
 				currentSection = {
-					name: section.name,
 					mode: SectionType.DepCache,
 					filters: section.modules.slice()
 				};
--- a/lib/lbt/bundle/AutoSplitter.js
+++ b/lib/lbt/bundle/AutoSplitter.js
@@ -41,6 +41,7 @@ class AutoSplitter {
 		const numberOfParts = options.numberOfParts;
 		let totalSize = 0;
 		const moduleSizes = Object.create(null);
+		const depCacheSymbol = Symbol("depCache");
 		this.optimize = !!options.optimize;
 
 		// ---- resolve module definition
@@ -75,7 +76,8 @@ class AutoSplitter {
 				});
 				break;
 			case SectionType.DepCache:
-				moduleSizes["__depCacheSize"] = "sap.ui.loader.config({depCacheUI5:{}});".length;
+				moduleSizes[depCacheSymbol] = "sap.ui.loader.config({depCacheUI5:{}});".length;
+				totalSize += moduleSizes[depCacheSymbol];
 
 				section.modules.forEach( (module) => {
 					promises.push((async () => {
@@ -87,7 +89,7 @@ class AutoSplitter {
 						);
 						if (deps.length > 0) {
 							const depSize = `"${module}": [${deps.map((dep) => `"${dep}"`).join(",")}],`.length;
-							moduleSizes["__depCacheSize"] += depSize;
+							moduleSizes[depCacheSymbol] += depSize;
 							totalSize += depSize;
 						}
 					})());
@@ -206,7 +208,7 @@ class AutoSplitter {
 					filters: section.modules.slice()
 				};
 				currentModule.sections.push( currentSection );
-				totalSize += moduleSizes["__depCacheSize"];
+				totalSize += moduleSizes[depCacheSymbol];
 				break;
 			default:
 				break;
--- a/lib/lbt/bundle/AutoSplitter.js
+++ b/lib/lbt/bundle/AutoSplitter.js
@@ -74,6 +74,25 @@ class AutoSplitter {
 					totalSize += "sap.ui.requireSync('');".length + toRequireJSName(module).length;
 				});
 				break;
+			case SectionType.DepCache:
+				moduleSizes["__depCacheSize"] = "sap.ui.loader.config({depCacheUI5:{}});".length;
+
+				section.modules.forEach( (module) => {
+					promises.push((async () => {
+						const resource = await this.pool.findResourceWithInfo(module);
+						const deps = resource.info.dependencies.filter(
+							(dep) =>
+								!resource.info.isConditionalDependency(dep) &&
+								!resource.info.isImplicitDependency(dep)
+						);
+						if (deps.length > 0) {
+							const depSize = `"${module}": [${deps.map((dep) => `"${dep}"`).join(",")}],`.length;
+							moduleSizes["__depCacheSize"] += depSize;
+							totalSize += depSize;
+						}
+					})());
+				});
+				break;
 			default:
 				break;
 			}
@@ -179,6 +198,15 @@ class AutoSplitter {
 					currentSection.filters.push( module );
 					totalSize += 21 + toRequireJSName(module).length;
 				});
+				break;
+			case SectionType.DepCache:
+				currentSection = {
+					name: section.name,
+					mode: SectionType.DepCache,
+					filters: section.modules.slice()
+				};
+				currentModule.sections.push( currentSection );
+				totalSize += moduleSizes["__depCacheSize"];
 				break;
 			default:
 				break;
--- a/lib/lbt/bundle/AutoSplitter.js
+++ b/lib/lbt/bundle/AutoSplitter.js
@@ -92,10 +92,7 @@ class AutoSplitter {
 							const depSize = `"${module}": [${deps.map((dep) => `"${dep}"`).join(",")}],`.length;
 							totalSize += depSize;
 
-							depCacheSizes.push({
-								size: depSize,
-								filters: [module, ...deps]
-							});
+							depCacheSizes.push({size: depSize, module});
 						}
 					})());
 				});
@@ -230,10 +227,10 @@ class AutoSplitter {
 						splittedModules.push(currentModule);
 					}
 
-					depCache.filters
-						.filter((moduleName) => !currentSection.filters.includes(moduleName))
-						.forEach((moduleName) => currentSection.filters.push(moduleName));
-					totalSize += depCache.size;
+					if (!currentSection.filters.includes(depCache.module)) {
+						currentSection.filters.push(depCache.module);
+						totalSize += depCache.size;
+					}
 				});
 				break;
 			default:
--- a/lib/lbt/bundle/AutoSplitter.js
+++ b/lib/lbt/bundle/AutoSplitter.js
@@ -41,7 +41,9 @@ class AutoSplitter {
 		const numberOfParts = options.numberOfParts;
 		let totalSize = 0;
 		const moduleSizes = Object.create(null);
-		const depCacheSymbol = Symbol("depCache");
+		const depCacheSizes = [];
+		let depCacheLoaderSize = 0;
+		let depCacheSet = new Set();
 		this.optimize = !!options.optimize;
 
 		// ---- resolve module definition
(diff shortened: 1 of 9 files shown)


lib/tasks/bundlers/generateLibraryPreload.js
```diff
@@ -31,11 +31,63 @@ function getDefaultLibraryPreloadFilters(namespace, excludes) {
 	return filters;
 }
 
-function getBundleDefinition(namespace, excludes) {
+function getBundleDefinition(namespace, excludes, preloadMode = "preload") {
```
matz3, line 34: There should not be any changes to the current way how library preloads are created. The legacy h2-preload should not be re-introduced. First step is only to make the new bundle section mode available.

lib/lbt/bundle/BundleDefinition.js
```diff
@@ -27,5 +27,13 @@ export const SectionType = {
...
 	 * execution of some specific module or modules.
 	 */
-	Require: "require"
+	Require: "require",
+
+	/**
+	 * Dependency cache information that lists modules and their dependencies
+	 * of all types: JS, declarative views/fragments.
+	 * Only the dependencies of the modules are stored as 'depCache' configuration.
+	 * Requires UI5 Evolution runtime.
```
matz3, line 36: Can be removed. It requires ui5loader (since UI5 1.54), which also applies to all other APIs currently used. Legacy API support has been removed with v3.0 (see https://github.com/SAP/ui5-builder/pull/739).
```suggestion
```

lib/lbt/bundle/BundleDefinition.js
```diff
@@ -27,5 +27,13 @@ export const SectionType = {
 	 * Usually used as the last section in a merged module to enforce loading and
 	 * execution of some specific module or modules.
 	 */
-	Require: "require"
+	Require: "require",
+
+	/**
+	 * Dependency cache information that lists modules and their dependencies
+	 * of all types: JS, declarative views/fragments.
+	 * Only the dependencies of the modules are stored as 'depCache' configuration.
```
matz3, line 35: Isn't this already clear with the first sentence? To me this sounds like duplicate information or is confusing.

lib/lbt/bundle/Builder.js
```diff
@@ -220,6 +228,8 @@ class BundleBuilder {
 			return this.writeBundleInfos([section]);
 		case SectionType.Require:
 			return this.writeRequires(section);
+		case SectionType.DepCache:
```
matz3, line 231: The AutoSplitter also needs to know about the DepCache SectionType. Otherwise it fails to calculate the correct bundle size in order to decide which parts of the bundle to put into separate files. The goal of the AutoSplitter is to split bundle content evenly across multiple parts (option: numberOfParts > 1).

It also seems that this has been missed to be implemented for the BundleInfo mode.

lib/lbt/bundle/BundleDefinition.js
d3xter666: Removed it

codeworrior: Maybe this was meant to explain that conditional dependencies are not excluded? if so, I would rather state it like that.

lib/lbt/bundle/Builder.js
```diff
@@ -549,6 +559,47 @@ class BundleBuilder {
 		});
 	}
 
+	async writeDepCache(section) {
+		const outW = this.outW;
+
+		outW.ensureNewLine();
```
codeworrior, line 565: maybe also move into the `if`

lib/lbt/bundle/Builder.js
```diff
@@ -549,6 +559,47 @@ class BundleBuilder {
...
+					if (deps.length > 0) {
+						if (i > 0) {
+							outW.writeln(",");
+						}
+						outW.write(
+							`"${module}": [${deps.map((dep) => `"${dep}"`).join(",")}]`
+						);
+						i++;
+					} else {
+						log.verbose("    skipped %s, no dependencies", module);
```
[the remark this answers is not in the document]
d3xter666: Done

lib/lbt/bundle/Builder.js
```diff
@@ -549,6 +559,47 @@ class BundleBuilder {
...
+					}
+				} else {
+					log.error("    couldn't find %s", module);
```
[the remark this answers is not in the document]
d3xter666: Thanks! Fixed it

lib/lbt/bundle/Builder.js
d3xter666: Done

lib/lbt/bundle/BundleDefinition.js
matz3: > Maybe this was meant to explain that conditional dependencies are not excluded? if so, I would rather state it like that.

Ah yeah, that makes sense 👍🏻

lib/lbt/bundle/Builder.js
d3xter666: Thanks!

test/lib/lbt/bundle/Builder.js
```diff
@@ -817,6 +817,104 @@ ${SOURCE_MAPPING_URL}=library-preload.js.map
 	]);
 });
 
+test.serial("integration: createBundle with depCache", async (t) => {
```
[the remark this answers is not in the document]
d3xter666: I have added a test where modules do not have dependencies, therefore the `sap.ui.loader.config({depCacheUI5: ....` won't be written at all. 
Was that your idea?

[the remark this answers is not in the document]
matz3: Yes, exactly.

lib/lbt/bundle/BundleDefinition.js
matz3: The conditional dependencies mentioned by @codeworrior are not internal, as they are reflected in the resources.json (`condRequired`). It would be helpful to mention that conditional dependencies are NOT listed in the depCache information.

test/lib/lbt/bundle/Builder.js
```diff
@@ -817,6 +817,198 @@ ${SOURCE_MAPPING_URL}=library-preload.js.map
...
+		"should contain:" +
+		" preload part from a.js" +
+		" depCache part from a.js && c2.js");
+	t.is(oResult.bundleInfo.name, "library-depCache-preload.js", "bundle info name is correct");
+	t.deepEqual(oResult.bundleInfo.size, expectedContent.length, "bundle info size is correct");
+	t.deepEqual(oResult.bundleInfo.subModules, ["a.js", "b.js", "c2.js", "c1.js", "c3.js"],
+		"bundle info subModules are correct");
+});
+
+test.serial("integration: createBundle with depCache with NO dependencies", async (t) => {
```
matz3, line 918: How is this test different to the other one? I think I'm missing something, but the bundle definition is exactly the same, isn't it?

lib/lbt/bundle/Builder.js
```diff
@@ -549,6 +559,50 @@ class BundleBuilder {
...
 	}
 
+	async writeDepCache(section) {
+		const outW = this.outW;
+		let hasDepCache = false;
+
+		const sequence = section.modules.slice().sort();
+
+		if (sequence.length > 0) {
+			// let i = 0;
```
matz3, line 569: remove if it's not needed anymore

d3xter666: Ah, thanks!

test/lib/lbt/bundle/Builder.js
d3xter666: Yes, the bundleDefinition is absolutely the same, but the resources are not.
Please take a look at this one, for example: https://github.com/SAP/ui5-builder/pull/951/files/798db531a87bda0e1ef37c779051a3fc934f9984#diff-a0b32670fda103e5bc38b7bb79742c5acf2b80edad872a32b112a23284fdd7beR926

None of the resources depends on any of the other resources.
The builder would go and analyse them, but they'll be skipped as they do not have any dependencies defined.

lib/lbt/bundle/AutoSplitter.js
```diff
@@ -74,6 +74,25 @@ class AutoSplitter {
 					totalSize += "sap.ui.requireSync('');".length + toRequireJSName(module).length;
 				});
 				break;
+			case SectionType.DepCache:
+				moduleSizes["__depCacheSize"] = "sap.ui.loader.config({depCacheUI5:{}});".length;
```
RandomByte, line 78: Maybe it makes sense to use a symbol here to prevent clashes with module names. I guess it's unlikely to happen but probably a good practice nevertheless.


```js
const depCacheSymbol = Symbol("depCache")
moduleSizes[depCacheSymbol] = ...
```

RandomByte, line 78: Also, don't we need to add the length to `totalSize` here as well (like in line 91)

test/lib/lbt/bundle/Builder.js
```diff
@@ -817,6 +817,198 @@ ${SOURCE_MAPPING_URL}=library-preload.js.map
 	]);
 });
 
+test.serial("integration: createBundle with depCache", async (t) => {
```
RandomByte, line 820: Missing test for the AutoSplitter changes (also see [code coverage](https://coveralls.io/builds/63974262/source?filename=lib%2Flbt%2Fbundle%2FAutoSplitter.js#L78))

test/lib/lbt/bundle/Builder.js
matz3: Ah got it, thanks.

lib/lbt/bundle/AutoSplitter.js
d3xter666: Done

d3xter666: Ah, yes! Thanks a lot!

lib/lbt/bundle/AutoSplitter.js
```diff
@@ -180,6 +199,15 @@ class AutoSplitter {
...
 				});
 				break;
+			case SectionType.DepCache:
+				currentSection = {
+					name: section.name,
+					mode: SectionType.DepCache,
+					filters: section.modules.slice()
+				};
+				currentModule.sections.push( currentSection );
+				totalSize += moduleSizes["__depCacheSize"];
```
codeworrior: The use of a fixed, single entry in `moduleSizes` limits us to a single `depCache` section and does not allow to split that section, if it is huge. (one section in the bundle definition could be split into multiple `sap.ui.loader.config` calls in the final bundle).

test/lib/lbt/bundle/Builder.js
d3xter666: Done

lib/lbt/bundle/AutoSplitter.js
d3xter666: Thanks, @codeworrior !
If I understand you correctly, this would mean that we can split the depCache definitions into different files, like the preload does. So, in the end all those definintions `sap.ui.loader.config({depCacheUI5:{}});` would be merged and not override themselves, is this correct?

RandomByte (commented):
LGTM

d3xter666:
> The depCache is currently broken when using the AutoSplitter (`numberOfParts > 1`). This can be easily reproduced within openui5-sample-app:
> 
> ```yaml
> builder:
>   bundles:
>     - bundleDefinition:
>         name: depCache-test-bundle.js
>         sections:
>           - filters:
>               - "sap/ui/demo/todo/"
>               - "sap/ui/demo/todo/manifest.json"
>             mode: depCache
>       bundleOptions:
>         numberOfParts: 2
> ```
> 
> The bundle then not only contains the modules from the filters (expected), but also dependencies of those modules (outside of the project), although `resolve` should default to `false`.
> 
> [0ecfd7d](https://github.com/SAP/ui5-builder/commit/0ecfd7dced28f8f0e749e506b3675904a4585a3e) (Enable auto split for depCache) is the commit which broke that behaviour.
> 
> A smaller issue is that the part sizes differ a lot (5kb vs 9kb).

Hi @matz3,

Thanks for finding those!
I have fixed the filtering for the modules.

Regarding the part sizes, it used to be due to the issue with the filtering above.
Still, there would be slight differences- the same as with `preload`s. It's not possible to have them perfectly equal.

Cheers

RandomByte (approved):

PR #953, opened 2023-11-08
[FIX] Handle graceful termination of workerpool for parallel builds

d3xter666:
fixes: https://github.com/SAP/ui5-tooling/issues/894
JIRA: CPOUI5FOUNDATION-751
depends on: https://github.com/SAP/ui5-project/pull/677

Workerpool needs to wait for all the active tasks to complete before terminating.

### DIFF
--- a/lib/processors/minifier.js
+++ b/lib/processors/minifier.js
@@ -39,9 +39,12 @@ function getPool(taskUtil) {
 
 				// There are many stats that could be used, but these ones seem the most
 				// convenient. When all the (available) workers are idle, then it's safe to terminate.
-				const {idleWorkers, totalWorkers} = pool.stats();
+				let {idleWorkers, totalWorkers} = pool.stats();
 				while (idleWorkers !== totalWorkers && !force) {
 					await setTimeoutPromise(100); // Wait a bit workers to finish and try again
+					const stats = pool.stats();
+					idleWorkers = stats.idleWorkers;
+					totalWorkers = stats.totalWorkers;
 				}
 
 				const poolToBeTerminated = pool;
--- a/lib/processors/minifier.js
+++ b/lib/processors/minifier.js
@@ -42,9 +42,7 @@ function getPool(taskUtil) {
 				let {idleWorkers, totalWorkers} = pool.stats();
 				while (idleWorkers !== totalWorkers && !force) {
 					await setTimeoutPromise(100); // Wait a bit workers to finish and try again
-					const stats = pool.stats();
-					idleWorkers = stats.idleWorkers;
-					totalWorkers = stats.totalWorkers;
+					({idleWorkers, totalWorkers} = pool.stats());
 				}
 
 				const poolToBeTerminated = pool;
--- a/lib/processors/minifier.js
+++ b/lib/processors/minifier.js
@@ -40,9 +40,8 @@ function getPool(taskUtil) {
 				// There are many stats that could be used, but these ones seem the most
 				// convenient. When all the (available) workers are idle, then it's safe to terminate.
 				const {idleWorkers, totalWorkers} = pool.stats();
-				if (idleWorkers !== totalWorkers && !force) {
+				while (idleWorkers !== totalWorkers && !force) {
 					await setTimeoutPromise(100); // Wait a bit workers to finish and try again
-					return attemptPoolTermination();
 				}
 
 				const poolToBeTerminated = pool;
--- a/lib/processors/minifier.js
+++ b/lib/processors/minifier.js
@@ -28,14 +28,20 @@ function getPool(taskUtil) {
 			workerType: "auto",
 			maxWorkers
 		});
-		taskUtil.registerCleanupTask(() => {
+		taskUtil.registerCleanupTask((options) => {
 			const attemptPoolTermination = () => {
 				log.verbose(`Attempt to terminate the workerpool...`);
+
+				if (!pool) {
+					return;
+				}
+
+				const forceTerminate = !!options && !!options.exitCode;
 
 				// There are many stats that could be used, but these ones seem the most
 				// convenient. When all the (available) workers are idle, then it's safe to terminate.
 				const {idleWorkers, totalWorkers} = pool.stats();
-				if (idleWorkers !== totalWorkers) {
+				if (idleWorkers !== totalWorkers && !forceTerminate) {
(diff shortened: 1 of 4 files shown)


lib/processors/minifier.js
```diff
@@ -29,10 +29,24 @@ function getPool(taskUtil) {
 			maxWorkers
 		});
 		taskUtil.registerCleanupTask(() => {
-			log.verbose(`Terminating workerpool`);
-			const poolToBeTerminated = pool;
-			pool = null;
-			poolToBeTerminated.terminate();
```
RandomByte, line 35: According to the [documentation](https://github.com/josdejong/workerpool#pool) `Pool.terminate` supposedly takes care of the graceful termination of workers that are still busy:

> If parameter `force` is false (default), workers will finish the tasks they are working on before terminating themselves. Any pending tasks will be rejected with an error 'Pool terminated'.

I actually wasn't aware that `Pool.terminate` returns a promise. Apparently the examples in the [README](https://github.com/josdejong/workerpool#use) don't wait for it either. But maybe that's the main fix here? Make the callback function async (taskUtil will wait for any returned promise) and  `await poolToBeTerminated.terminate();`?

RandomByte (commented):
Just to get a clear picture: Can the issue being fixed here occur in regular UI5 CLI builds? Or only in scenarios like https://github.com/SAP/ui5-tooling/issues/894 where multiple minify tasks run in parallel?

lib/processors/minifier.js
d3xter666: Thank you, @RandomByte!
Basically, it would make the issue harder to reproduce, but it would still be there.
Here the key point is `tasks they are working on`. 
Your proposal makes a lot of sense IF `busy workers` === `total tasks`.

Please take a look at the tests cases with this change. I intentionally made them with more data to expose the real problem. We need to wait until there are no busy, pending nor active tasks in the queue. Awaiting pool's termination would wait for the active tasks to finnish, but all of the pending tasks would be cancelled.

RandomByte: Indeed I missed how workerpool handles queued task during its termination. Since it **[rejects]()** all queued tasks, we end up with `Pool terminated` issue in scenarios like https://github.com/SAP/ui5-tooling/issues/894

So this change fixes that case by letting all queued tasks finish executing before terminating the pool.

However, this means when *cancelling* a UI5 CLI build (e.g. using CTRL+C), the pool will also wait for all queued tasks to finish before terminating. This can be noticed for example when building the `sap.m` library and cancelling the build right after the `minify` task starts. With this PR it takes ~2 seconds for ending the process. Before it would end almost immediately

d3xter666: Thanks for finding it out!
We'll need to have the execution contex in that case. Indeed, it's quite slow, when we forcefully try to terminate it. In this case it should enforce the termination.

I have created a complimentary change that provides this context: https://github.com/SAP/ui5-project/pull/677

RandomByte (approved):
LGTM but depends on https://github.com/SAP/ui5-project/pull/677

lib/processors/minifier.js
```diff
@@ -28,11 +28,29 @@ function getPool(taskUtil) {
...
+				if (!pool) {
+					return;
+				}
+
+				// There are many stats that could be used, but these ones seem the most
+				// convenient. When all the (available) workers are idle, then it's safe to terminate.
+				const {idleWorkers, totalWorkers} = pool.stats();
+				if (idleWorkers !== totalWorkers && !force) {
+					return new Promise((resolve) =>
+						setTimeout(() => resolve(attemptPoolTermination()), 100) // Retry after a while
```
matz3, line 44: There's no error propagation when `attemptPoolTermination()` fails synchronously, as the `reject` function of the `new Promise` is not called.
Can't you just use the promise variant of setTimeout here (`node:timers/promises`)?
The recursion might even be replaced with a loop (I might be wrong about that).

d3xter666: Thanks!
Indeed, I was able to remove the recursion that way. Seems more clear now.

lib/tasks/buildThemes.js
```diff
@@ -34,10 +35,8 @@ function getPool(taskUtil) {
 				// There are many stats that could be used, but these ones seem the most
 				// convenient. When all the (available) workers are idle, then it's safe to terminate.
 				const {idleWorkers, totalWorkers} = pool.stats();
```
matz3, line 37: Are those properties updated automatically, or how does the loop work without calling `pool.stats()` again?

d3xter666: Yep, sorry, didn't sync with the latest change.
Thanks!

lib/processors/minifier.js
```diff
@@ -39,9 +39,12 @@ function getPool(taskUtil) {
 
 				// There are many stats that could be used, but these ones seem the most
 				// convenient. When all the (available) workers are idle, then it's safe to terminate.
-				const {idleWorkers, totalWorkers} = pool.stats();
+				let {idleWorkers, totalWorkers} = pool.stats();
 				while (idleWorkers !== totalWorkers && !force) {
 					await setTimeoutPromise(100); // Wait a bit workers to finish and try again
+					const stats = pool.stats();
```
RandomByte, line 45:
```suggestion
					({idleWorkers, totalWorkers} = pool.stats());
```

lib/tasks/buildThemes.js
```diff
@@ -34,9 +34,12 @@ function getPool(taskUtil) {
 
 				// There are many stats that could be used, but these ones seem the most
 				// convenient. When all the (available) workers are idle, then it's safe to terminate.
-				const {idleWorkers, totalWorkers} = pool.stats();
+				let {idleWorkers, totalWorkers} = pool.stats();
 				while (idleWorkers !== totalWorkers && !force) {
 					await setTimeoutPromise(100); // Wait a bit workers to finish and try again
+					const stats = pool.stats();
```
RandomByte, line 40:
```suggestion
					({idleWorkers, totalWorkers} = pool.stats());
```

lib/tasks/buildThemes.js
RandomByte: I guess the tests didn't fail because one 100ms delay is already enough to clear the queue?

matz3 (approved):

RandomByte (approved):

PR #954, opened 2023-11-08
[FIX] Bundling: Detect manifest.json dependency of libraries

matz3:
This solves the problem of missing manifest.json files of libraries in
bundles, such as sap-ui-custom.js (self-contained build).

It also eliminates the need to explicitly mention manifest.json files
of libraries to be included in custom bundles.


RandomByte:
LGTM but I think @codeworrior should have a look too

RandomByte (approved):
LGTM

PR #957, opened 2023-11-20
[FEATURE] manifest.json: Auto-fill supportedLocales

flovogt:
After the UI5 Tooling build, the `supportedLocales` property in the `manifest.json` of a UI5 library/application will be filled automatically with the available locales determined by the existence of the respective `.properties` file.

To disable the automatic generation of the `supportedLocales`, set `supportedLocales` to any desired value.

### Scenario: Application

````txt
- webapp/i18n/
  - i18n.properties
  - i18n_en.properties
  - i18n_en_US.properties
  - i18n_de.properties
  - i18n_de_DE.properties 
````
In the `manifest.json` the `supportedLocales` property will be enhanced as following:

#### Source
`````json
"models": {
  "i18n": {
    "type": "sap.ui.model.resource.ResourceModel",
    "settings": {
	    "bundleName": "sap.ui.demo.todo.i18n.i18n"
    }
 }
`````

#### Build Result
`````json
"models": {
  "i18n": {
    "type": "sap.ui.model.resource.ResourceModel",
    "settings": {
	    "bundleName": "my.app.i18n.i18n"
	    "supportedLocales": [
                 "",
                  "de",
                  "de_DE",
                  "en",
                  "en_US"
	    ]
    }
 }
`````

JIRA: CPOUI5FOUNDATION-296

Depends on:
- https://github.com/SAP/ui5-server/pull/617
- https://github.com/SAP/ui5-project/pull/683

Demo: https://github.com/SAP/openui5-sample-app/pull/220

### CODE REVIEW

File: lib/processors/manifestTransformer.js [hunk 1]
```diff
@@ -0,0 +1,218 @@
...
+	return bundleConfigs.concat(Object.values(sapui5ModelConfig)
+		.filter((model) => model.type === "sap.ui.model.resource.ResourceModel")
+		.map((model) => model.settings));
+}
+
+function getSapUi5LibrariesBundles(manifest, bundleConfigs) {
+	const sapui5LibraryConfig = manifest?.["sap.ui5"]?.library;
+	let sapui5LibraryBundleConfig = sapui5LibraryConfig?.i18n;
+
+	if (!sapui5LibraryBundleConfig) {
```
matz3, line 99: Runtime code in sap/ui/core/Lib#_normalizeI18nSettings says:
> true|null|undefined - use default settings: bundle is 'messageBundle.properties'

lib/processors/manifestTransformer.js
```diff
@@ -0,0 +1,218 @@
...
+ * Transforms the content of the manifest.json file.
+ *
+ * @public
+ * @function default
+ * @static
+ *
+ * @param {object} parameters Parameters
+ * @param {@ui5/fs/Resource[]} parameters.resources List of manifest.json resources to be processed
+ * @param {fs|module:@ui5/fs/fsInterface} parameters.fs Node fs or custom
+ *    [fs interface]{@link module:@ui5/fs/fsInterface}. Required when setting "readSourceMappingUrl" to true
```
matz3, line 212: Copy & paste?

lib/processors/manifestTransformer.js
```diff
@@ -0,0 +1,218 @@
...
+		let propertyFilesPath = bundleUrl;
+
+		if (bundleName) {
+			propertyFilesPath = bundleName.replace(/\./g, "/") + ".properties";
+		}
+
+		const propertyFileName = propertyFilesPath.substring(propertyFilesPath.lastIndexOf("/") + 1);
+		propertyFilesPath = propertyFilesPath.substring(0, propertyFilesPath.lastIndexOf("/"));
+
+		const generatedSupportedLocales = fs.readdir(`/resources/${propertyFilesPath}/!(${propertyFileName}`)
```
matz3, line 173: fs.readdir is async and accepts a callback. It's the default Node.js "fs" interface. It usually makes sense to use util.promisify to get a promise version of a function.

lib/processors/manifestTransformer.js
```diff
@@ -0,0 +1,218 @@
...
+
+	return true;
+}
+
+function getType(manifest) {
+	return manifest["sap.app"].type;
+}
+
+function isFallbackLocalePartOfSupportedLocales(fallbackLocale, supportedLocales) {
+	return supportedLocales.indexOf(fallbackLocale) > -1;
```
matz3, line 58: includes?

lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,313 @@
...
+}
+
+/**
+ * Transforms "ui5://my/fancy/app/i18n/i18n.properties" to "my.fancy.app.i18n.i18n"
+ *
+ * @param {string} bundleUrl A bundle Url
+ * @returns {string}
+ */
+function getBundleNameByUrl(bundleUrl) {
+	return bundleUrl.split("ui5://")[1].replaceAll("/", ".").split(".properties")[0];
```
d3xter666, line 48: Is it possible that `ui5://` could be missing?
Maybe a `.replace("ui5://", "")` would be safer

lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,313 @@
...
+
+function isFallbackLocalePartOfSupportedLocales(fallbackLocale, supportedLocales) {
+	return supportedLocales.includes(fallbackLocale);
+}
+
+function hasBundleTerminologies(bundleConfig) {
+	return bundleConfig?.terminologies && Object.keys(bundleConfig.terminologies).length > 0;
+}
+
+function getTerminologieBundles(bundleConfig, bundleConfigs) {
```
d3xter666, line 104: The name suggests that this should be a getter, but instead modifies `bundleConfigs`

lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,313 @@
...
+	const terminologyBundles = [];
+	if (hasBundleTerminologies(bundleConfig)) {
+		Object.keys(bundleConfig.terminologies).forEach((key) => {
+			terminologyBundles.push(bundleConfig.terminologies[key]);
+		});
+	}
+	bundleConfigs.push(...terminologyBundles);
+}
+
+function getEnhanceWithBundles(bundleConfig, bundleConfigs) {
```
d3xter666, line 114: The name suggests that this should be a getter, but instead modifies `bundleConfigs`.

lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	bundleConfig.enhanceWith.forEach((config) => {
+		getTerminologieBundles(config, bundleConfigs);
+	});
+
+	bundleConfigs.push(...bundleConfig.enhanceWith);
+}
+
+function getSapAppBundle(manifest, bundleConfigs) {
```
d3xter666, line 126: The name suggests that this should be a getter, but instead modifies `bundleConfigs`.

lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,313 @@
...
+	} else {
+		getTerminologieBundles(sapAppBundleConfig, bundleConfigs);
+		getEnhanceWithBundles(sapAppBundleConfig, bundleConfigs);
+	}
+
+	sapAppConfig.i18n = sapAppBundleConfig;
+	bundleConfigs.push(sapAppBundleConfig);
+}
+
+function getSapUi5ModelBundles(manifest, bundleConfigs) {
```
d3xter666, line 157: The name suggests that this should be a getter, but instead modifies `bundleConfigs`.

lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,313 @@
...
+	const descriptorVersion = new Version(manifest._version);
+	if (descriptorVersion.compare(APP_DESCRIPTOR_V22) === -1) {
+		log.verbose("manifest.json: _version is lower than 1.21.0 so no supportedLocales can be generated");
+		return;
+	}
+
+	const project = resource.getProject();
+	const namespace = project.getNamespace();
+
+	// in case of usage by server middleware --> isNamespaced ist false
```
d3xter666, line 233:
```suggestion
	// in case of usage by server middleware --> isNamespaced is false
```

lib/tasks/enrichManifest.js
```diff
@@ -0,0 +1,41 @@
...
+ * @param {@ui5/fs/DuplexCollection} parameters.workspace DuplexCollection to read and write files
+ * @param {object} parameters.options Options
+ * @param {string} parameters.options.projectNamespace Namespace of the application
+ * @returns {Promise<undefined>} Promise resolving with <code>undefined</code> once data has been written
+ */
+export default async function({workspace, options}) {
+	// Backward compatibility: "namespace" option got renamed to "projectNamespace"
+	const namespace = options.projectNamespace || options.namespace;
+
+	// Note: all "manifest.json" files
```
d3xter666, line 24: in this namespace?

lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,313 @@
...
+ * Determines if manifest contain handlebars placeholder templates
+ *
+ * Only in case of placeholders in manifest.json the bundle configured
+ * in sap.app/i18n is requested by sap/ui/core/Manifest.js
+ *
+ * @param {object} manifest The manifest
+ * @returns {boolean}
+ */
+function hasManifestTemplates(manifest) {
+	const rManifestTemplate = /\{\{([^}}]+)\}\}/g;
```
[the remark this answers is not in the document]
flovogt: I used the same regex like in openui5. I would like to have both places in sync.

lib/processors/manifestEnricher.js
flovogt: changed to `replaceAll` also last split in this line

lib/processors/manifestEnricher.js
flovogt: done

lib/tasks/enrichManifest.js
flovogt: done

lib/processors/manifestTransformer.js
```diff
@@ -0,0 +1,218 @@
...
+
+function getSapAppBundle(manifest, bundleConfigs) {
+	const sapAppConfig = manifest["sap.app"];
+	let sapAppBundleConfig = sapAppConfig?.i18n;
+
+	if (getType(manifest) !== "library") {
+		if (hasManifestTemplates(manifest)) {
+			// sap/ui/core/Manifest defaults sap.app/i18n to "i18n/i18n.properties"
+			// when template is used in manifest.json
+			sapAppBundleConfig = sapAppBundleConfig || "i18n/i18n.properties";
```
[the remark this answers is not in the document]
flovogt: yes bundle is only loaded when the manifest.json contains a template e.g `{{TITLE}}` and this is checked by in the processor too

lib/processors/manifestTransformer.js
flovogt: valid point.... `undefined`/ `null` should be treated like `true`. I will adjust the code and tests

flovogt: Runtime says 

```
true|null|undefined - use default settings: bundle is 'messageBundle.properties',
fallback and supported locales are not defined (defaulted by ResourceBundle)
```
So either we skip the generation of supportedLocales when value is `true`, `null` or `undefined`
or we generate them but then the runtime comment is no longer correct. @matz3 whats your opinion on that one?

matz3: Can't we convert `true|null|undefined` to an object `{bundleUrl: "messagebundle.properties"}` and then add the `supportedLocales` ? That conversion would happen at runtime anyways, and we just add the `supportedLocales` information. Of course, this default must be aligned with the runtime code, but I doubt that it will change any time in future.

lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,313 @@
...
+ *
+ * @public
+ * @function default
+ * @static
+ *
+ * @param {object} parameters Parameters
+ * @param {@ui5/fs/Resource[]} parameters.resources List of manifest.json resources to be processed
+ * @param {fs|module:@ui5/fs/fsInterface} parameters.fs Node fs or custom
+ *    [fs interface]{@link module:@ui5/fs/fsInterface}.
+ * @param {object} parameters.options Options
```
matz3, line 308: Are there any options? If so, please document them.

lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,313 @@
...
+ * @returns {Promise<@ui5/fs/Resource[]>} Promise resolving with the cloned resources
```
matz3, line 309: Does it really resolve with a cloned resource? I don't see that in the code and I think it's fine if we adjust the existing resource via setString.

lib/tasks/enrichManifest.js
```diff
@@ -0,0 +1,41 @@
...
+ * @static
+ *
+ * @param {object} parameters Parameters
+ * @param {@ui5/fs/DuplexCollection} parameters.workspace DuplexCollection to read and write files
+ * @param {object} parameters.options Options
+ * @param {string} parameters.options.projectNamespace Namespace of the application
+ * @returns {Promise<undefined>} Promise resolving with <code>undefined</code> once data has been written
+ */
+export default async function({workspace, options}) {
+	// Backward compatibility: "namespace" option got renamed to "projectNamespace"
```
matz3, line 21: Do we have to add this to a new task? I would hope that it is only relevant for existing tasks.

lib/tasks/enrichManifest.js
```diff
@@ -0,0 +1,41 @@
...
+			return manifestEnricher({
+				resources,
+				fs: fsInterface(workspace),
+				options
+			});
+		})
+		.then((resources) => resources.map((resource) => {
+			if (resource) {
+				log.verbose("Resource transformed: " + resource.getPath());
+				workspace.write(resource);
```
matz3, line 36: `write` returns a Promise. Is this promise awaited? I don't see a Promise.all that handles this. Maybe `async / await` would also improve readability in this module.

lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,313 @@
...
+		return false;
+	}
+
+	if (bundleUrl && bundleUrl.startsWith("/")) {
+		log.verbose("manifest.json: bundleUrl '" + bundleUrl +
+			"' contains an absolute path, no supportedLocales are generated");
+		return false;
+	}
+
+	if (bundleUrl && bundleUrl.startsWith("..")) {
```
matz3, line 72: So one could workaround this limitation by just specifying `./../../`, right?
I highly recommend to properly resolve the url and then check whether the path is within the expected namespace. In this case it's not a big problem, but in other scenarios such checks could lead to security vulnerabilities, so I'd recommend to always ensure proper handling of input.

lib/processors/manifestTransformer.js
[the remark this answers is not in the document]
matz3: I don't feel like we should couple that default to the presence of a template. The default always applies. The runtime just doesn't do the "defaulting" if there's no template, as it doesn't need to load the ResourceBundle. But this might change in future. The runtime might in future decide to load the ResourceBundle for other purposes. So in the tooling, we should always consider this default, and add the `supportedLocales`.

lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,313 @@
...
+		} else if (!fallbackLocale && !isFallbackLocalePartOfSupportedLocales("en", generatedSupportedLocales)) {
+			log.warn("manifest.json: Generated supported locales ('" + generatedSupportedLocales.join("', '") + "') " +
+				"do not contain default fallback locale 'en'. Either provide a " +
+				"properties file for 'en' or configure another available fallbackLocale");
+		}
+
+		bundleConfig.supportedLocales = generatedSupportedLocales;
+	}));
+
+	resource.setString(JSON.stringify(manifest, null, options.prettyPrint ? "  " : undefined));
```
matz3, line 288: I would prefer to use tabs for formatting, as that's the recommendation by UI5.
Also, I don't see the need for a non-pretty-printed file, but I'm okay to keep this option as long as it defaults to true and is documented properly.

matz3 (changes requested):

test/lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,1647 @@
...
+			"type": "application"
+		},
+		"sap.ui5": {
+			"models": {
+				"i18n": {
+					"type": "sap.ui.model.resource.ResourceModel",
+					"settings": {
+						"bundleUrl": "../i18n/i18n.properties",
+						"fallbackLocale": "de",
+						"supportedLocales": ["de", "en"]
```
matz3: This expectation doesn't match with the description of the test. I think this is unexpected and the test should actually expect that the manifest is unchanged.

test/lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,1647 @@
...
+						"bundleName": "sap.ui.demo.app.i18n.i18n",
+						"fallbackLocale": "fr"
+					}
+				}
+			}
+		}
+	}, null, 2);
+
+	const resource = createResource("/resources/sap/ui/demo/app/manifest.json", true, input,
+		(actual) => t.deepEqual(actual, expected, "Correct file content should be set"));
```
matz3: I think setString should not be called, when the file is not expected to change.

lib/processors/manifestTransformer.js
matz3: I've update the PR and addressed this

lib/processors/manifestEnricher.js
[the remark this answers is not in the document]
matz3: IMO, this logic is not needed for our purposes. I've therefore removed the function.

lib/processors/manifestEnricher.js
matz3: I've updated the PR to rather call those methods `process...`, and they also modify the manifest, but only in case supportedLocales are added.

lib/processors/manifestEnricher.js
matz3: I've updated the PR to rather call those methods `process...`, and they also modify the manifest, but only in case supportedLocales are added.

lib/processors/manifestEnricher.js
matz3: I've updated the PR to rather call those methods `process...`, and they also modify the manifest, but only in case supportedLocales are added.

lib/processors/manifestEnricher.js
matz3: I've updated the PR to rather call those methods `process...`, and they also modify the manifest, but only in case supportedLocales are added.

lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,313 @@
...
+		.filter((model) => model.type === "sap.ui.model.resource.ResourceModel")
+		.map((model) => model.settings);
+	sapui5ModelConfigBundles.forEach((bundleConfig) => {
+		getTerminologieBundles(bundleConfig, bundleConfigs);
+		getEnhanceWithBundles(bundleConfig, bundleConfigs);
+	});
+	bundleConfigs.push(...sapui5ModelConfigBundles);
+}
+
+function getSapUi5LibrariesBundles(manifest, bundleConfigs) {
```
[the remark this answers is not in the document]
matz3: I've updated the PR to rather call those methods `process...`, and they also modify the manifest, but only in case supportedLocales are added.

test/lib/processors/manifestEnricher.js
```diff
@@ -0,0 +1,1647 @@
...
+
+
+function createResource(path, bNamespaced, input, fnOnSetString) {
+	return {
+		getString: () => Promise.resolve(input),
+		setString: fnOnSetString,
+		getProject() {
+			return {
+				getNamespace() {
+					const namespace = path.substring(0, path.lastIndexOf("/") + 1).replace("/resources/", "");
```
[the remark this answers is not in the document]
matz3: Done, as there is no explicit namespace handling anymore. Paths are resolved relative to the resource

test/lib/processors/manifestEnricher.js
matz3: Test has been updated accordingly

test/lib/processors/manifestEnricher.js
matz3: Test has been updated accordingly

matz3:
This still needs more work (refactoring, adding missing test cases)

lib/tasks/enrichManifest.js
```diff
@@ -0,0 +1,31 @@
...
+ * @param {string} parameters.options.projectNamespace Namespace of the application
+ * @returns {Promise<undefined>} Promise resolving with <code>undefined</code> once data has been written
+ */
+export default async function({workspace, options}) {
+	const {projectNamespace} = options;
+
+	// Note: all "manifest.json" files in the given namespace
+	const resources = await workspace.byGlob(`/resources/${projectNamespace}/**/manifest.json`);
+
+	const processedResources = await manifestEnricher({
```
RandomByte, line 24: After looking at this again and again I still stumble over the name manifest**Enricher**. Even though we agreed on this name a while ago, personally I think I would prefer a different name at this time.

### Definition of _enrichment_:
> the act or process of improving the quality or power of something by adding something else

This is indeed fitting, however the thing that does the enrichment is usually called "enrichment *something* (e.g. facility)" rather than "*something* enricher". See also [wiktionary](https://en.wiktionary.org/wiki/enricher)

### Definition for _augmentation_:

> the action or process of making or becoming greater in size or amount.

I find this fitting too and "[manifestAugmenter](https://en.wiktionary.org/wiki/augmenter)" works much better in my opinion.

### Definition for _enhancement_:

> an increase or improvement in quality, value, or extent.

"[manifestEnhancer](https://en.wiktionary.org/wiki/enhancer)"  also sounds better to me.


### Options

1. manifestEnricher (enrichManifest)
2. manifestAugmenter (augmentManifest)
3. manifestEnhancer (enhanceManifest)

flovogt: @KlattG any recommendations from your side. I am fine with all options mentioned by @RandomByte

KlattG: @flovogt I like enhancer best, augmenter second, and enricher would be my third choice.

lib/tasks/enrichManifest.js
flovogt: removed

lib/processors/manifestEnricher.js
flovogt: done

lib/tasks/enrichManifest.js
flovogt: rename done

lib/tasks/enhanceManifest.js
```diff
@@ -0,0 +1,31 @@
...
+export default async function({workspace, options}) {
+	const {projectNamespace} = options;
+
+	// Note: all "manifest.json" files in the given namespace
+	const resources = await workspace.byGlob(`/resources/${projectNamespace}/**/manifest.json`);
+
+	const processedResources = await manifestEnhancer({
+		resources,
+		fs: fsInterface(workspace),
+		options
```
RandomByte, line 27:
```suggestion
```

manifestEnhancer does not take in any options, so no need to pass them

lib/tasks/enhanceManifest.js
```diff
@@ -0,0 +1,31 @@
+import manifestEnhancer from "../processors/manifestEnhancer.js";
+import fsInterface from "@ui5/fs/fsInterface";
+
+/* eslint "jsdoc/check-param-names": ["error", {"disableExtraPropertyReporting":true}] */
+/**
+ * Task for transforming the manifest.json file.
```
RandomByte, line 6:
```suggestion
 * Task for transforming the manifest.json file.
 * Adds missing information based on the available project resources,
 * for example the locales supported by the present i18n resources.
```

lib/processors/manifestEnhancer.js
```diff
@@ -0,0 +1,417 @@
...
+			const manifest = await resource.getString();
+			const filePath = resource.getPath();
+			const manifestEnhancer = new ManifestEnhancer(manifest, filePath, fs);
+			const enrichedManifest = await manifestEnhancer.run();
+			if (enrichedManifest) {
+				resource.setString(JSON.stringify(enrichedManifest, null, 2));
+			}
+			return resource;
+		})
+	);
```
RandomByte, line 413: I think we should only return a resource if it has been modified. Maybe like this?

```suggestion
	const res = await Promise.all(
		resources.map(async (resource) => {
			const manifest = await resource.getString();
			const filePath = resource.getPath();
			const manifestEnhancer = new ManifestEnhancer(manifest, filePath, fs);
			const enrichedManifest = await manifestEnhancer.run();
			if (enrichedManifest) {
				resource.setString(JSON.stringify(enrichedManifest, null, 2));
				return resource;
			}
		})
	);
	return res.filter(($) => $);
```

lib/processors/manifestEnhancer.js
```diff
@@ -0,0 +1,417 @@
...
+				`bundleUrl '${bundleUrl}' points to a bundle outside of the ` +
+				`current namespace '${sapAppId}', enhancement of 'supportedLocales' is skipped`
+			);
+			return [];
+		}
+		const supportedLocales = await this.findSupportedLocales(normalizedBundleUrl);
+		if (!isTerminologyBundle && supportedLocales.length > 0) {
+			if (fallbackLocale && !supportedLocales.includes(fallbackLocale)) {
+				log.error(
+					`manifest.json: ` +
```
RandomByte, line 215: I think we should include the virtual path of the affected manifest.json to make it easier to identify which file we are complaining about.

```suggestion
					`${this.filePath}: ` +
```

lib/processors/manifestEnhancer.js
```diff
@@ -0,0 +1,417 @@
...
+					`Generated supported locales ('${supportedLocales.join("', '")}') for ` +
+					`bundle '${normalizedBundleUrl}' ` +
+					"not containing the defined fallback locale '" + fallbackLocale + "'. Either provide a " +
+					"properties file for defined fallbackLocale or configure another available fallbackLocale"
+				);
+				return [];
+			} else if (!fallbackLocale && !supportedLocales.includes("en")) {
+				log.warn(
+					`manifest.json: ` +
```
RandomByte, line 224:
```suggestion
					`${this.filePath}: ` +
```

lib/processors/manifestEnhancer.js
```diff
@@ -0,0 +1,417 @@
...
+class ManifestEnhancer {
+	/**
+	 * @param {string} manifest manifest.json content
+	 * @param {string} filePath manifest.json file path
+	 * @param {fs} fs Node fs or custom [fs interface]{@link module:@ui5/fs/fsInterface}
+	 */
+	constructor(manifest, filePath, fs) {
+		this.fsReadDir = promisify(fs.readdir);
+		this.cwd = path.dirname(filePath);
+		this.manifest = JSON.parse(manifest);
```
RandomByte, line 130: Add filePath so we can use it in logging (see below)

```suggestion
		this.filePath = filePath;
		this.manifest = JSON.parse(manifest);
```

lib/processors/manifestEnhancer.js
```diff
@@ -0,0 +1,417 @@
...
+		const resolvedBundleUrl = resolveUI5Url(bundleUrl);
+		if (!resolvedBundleUrl) {
+			// In case of a relative ui5-protocol URL
+			return [];
+		}
+		const sapAppId = this.manifest["sap.app"].id;
+		const normalizedBundleUrl = normalizeBundleUrl(resolvedBundleUrl, sapAppId);
+		if (normalizedBundleUrl.startsWith("../")) {
+			log.verbose(
+				`bundleUrl '${bundleUrl}' points to a bundle outside of the ` +
```
RandomByte, line 206:
```suggestion
				`${this.filePath}: ` +
				`bundleUrl '${bundleUrl}' points to a bundle outside of the ` +
```

lib/processors/manifestEnhancer.js
```diff
@@ -0,0 +1,417 @@
...
+
+	async run() {
+		// Prevent multiple invocations
+		if (this.runInvoked) {
+			throw new Error("ManifestEnhancer#run can only be invoked once per instance");
+		}
+		this.runInvoked = true;
+
+		if (!this.manifest._version) {
+			log.verbose("manifest.json: _version is not defined. No supportedLocales are generated");
```
RandomByte, line 360:
```suggestion
			log.verbose(`${this.filePath}: _version is not defined. No supportedLocales are generated`);
```

lib/processors/manifestEnhancer.js
```diff
@@ -0,0 +1,417 @@
...
+			return;
+		}
+
+		if (lt(this.manifest._version, APP_DESCRIPTOR_V22)) {
+			log.verbose("manifest.json: _version is lower than 1.21.0 so no supportedLocales can be generated");
```
RandomByte, line 365:
```suggestion
			log.verbose(`${this.filePath}: _version is lower than 1.21.0 so no supportedLocales can be generated`);
```

RandomByte (dismissed):
Just a few small findings

matz3 (commented):
LGTM

matz3 (approved):

matz3 (approved):

PR #958, opened 2023-11-23
[FIX] Incomplete string escaping or encoding

d3xter666:
JIRA: CPOUI5FOUNDATION-728
fixes: https://github.com/SAP/ui5-builder/security/code-scanning/2

### DIFF
--- a/lib/lbt/resources/ResourceFilterList.js
+++ b/lib/lbt/resources/ResourceFilterList.js
@@ -12,7 +12,7 @@ function makeFileTypePattern(fileTypes) {
 		if ( !type.startsWith(".") ) {
 			type = "." + type;
 		}
-		return type.replace(/[*+?.()|^$]/g, "\\$&");
+		return type.replace(/[*+?.()|^$\\]/g, "\\$&");
 	}).join("|") + ")";
 }
 


lib/lbt/resources/ResourceFilterList.js
```diff
@@ -12,7 +12,7 @@ function makeFileTypePattern(fileTypes) {
 		if ( !type.startsWith(".") ) {
 			type = "." + type;
 		}
-		return type.replace(/[*+?.()|^$]/g, "\\$&");
+		return type.replace(/[*+?.()|^$\\]/g, "\\$&");
```
codeworrior, line 15: Where do backslashes occur in our filters? We expect posix syntax.

d3xter666: Maybe, this change needs more context: https://github.com/SAP/ui5-builder/security/code-scanning/2

This CodeQL rule doesn't actually analyze the code, but does a static code check and as far as I got it, [this rule](https://github.com/github/codeql/blob/ff65ffafb003e86bd4c18f2c031f5faaf6f1038b/javascript/ql/src/Security/CWE-116/IncompleteSanitization.ql) checks for string methods that accept Regex as an argument (replace, replaceAll) and if the regex checks for special characters, it also (according to the rule) must escape a backslash.

So, we have several options here:
- ignore the rule as we know that we don't have backslashes in our input
- add a backslash to ensure the rule
- rewrite the replacement functionality in a way that does not use regex.

What's your opinion on that? Shall we just ignore it?

codeworrior: Thank you for the details, I should have checked the referenced BLI first :-)

As the escaping of "\" does not break anything ("\" should not occur in our file types), I don't mind escaping it.

codeworrior (approved):

PR #964, opened 2023-11-30
[INTERNAL] Align jsdoc template with openui5

codeworrior:


lib/processors/jsdoc/lib/transformApiJson.cjs
```diff
@@ -5,6 +5,8 @@
  * Licensed under the Apache License, Version 2.0 - see LICENSE.txt.
  */
 
+/* eslint-disable */
```
codeworrior, line 8: This is a bit unfortunate and you might not like it. But I would prefer to keep the files in sync as much as possible. 

One possible way out would be to make this an external config in openui5 (using a local .eslintrc.json for the lib or for the jsdoc folder) and remove the comment from this file here.

flovogt: Fine for me

flovogt (approved):

lib/processors/jsdoc/lib/transformApiJson.cjs
RandomByte: Fine with me too

RandomByte (approved):

PR #989, opened 2024-02-19
[BREAKING] Require Node.js 20.11.x/>=21.2.0 and npm >=10

RandomByte:
BREAKING CHANGE:
Support for older Node.js and npm releases has been dropped.
Only Node.js 20.11.x and >=21.2.0 as well as npm v10 or higher are supported.

### DIFF
--- a/lib/processors/jsdoc/jsdocGenerator.js
+++ b/lib/processors/jsdoc/jsdocGenerator.js
@@ -4,11 +4,7 @@ import path from "node:path";
 import {promisify} from "node:util";
 const writeFile = promisify(fs.writeFile);
 import {createAdapter} from "@ui5/fs/resourceFactory";
-import {createRequire} from "node:module";
 import {fileURLToPath} from "node:url";
-
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
-const require = createRequire(import.meta.url);
 
 /**
  * @public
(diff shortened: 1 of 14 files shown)


lib/processors/jsdoc/jsdocGenerator.js
```diff
@@ -93,7 +89,7 @@ async function generateJsdocConfig({targetPath, tmpPath, namespace, projectName,
 	const backslashRegex = /\\/g;
 
 	// Resolve path to this script to get the path to the JSDoc extensions folder
-	const jsdocPath = path.normalize(__dirname);
+	const jsdocPath = path.normalize(import.meta.dirname);
```
d3xter666, line 92: This one seems to be added in Node 21.2.1 https://nodejs.org/api/esm.html#importmetadirname and there's a chance we might fail in certain cases

RandomByte: See also the documentation pages:
v20: https://nodejs.org/docs/latest-v20.x/api/esm.html#importmetadirname
v21: https://nodejs.org/api/esm.html#importmetadirname

package.json
```diff
@@ -28,8 +28,8 @@
 		"./internal/jsdoc/template/publish": "./lib/processors/jsdoc/lib/ui5/template/publish.cjs"
 	},
 	"engines": {
-		"node": "^16.18.0 || >=18.12.0",
-		"npm": ">= 8"
+		"node": "^20.11.0 || >=21.2.0",
+		"npm": ">= 10"
```
matz3, line 32: Do we have a need to raise this? If it's technically not require from our side, I would rather not change it. Especially as npm 10 has some bugs and regressions.

RandomByte: Node v20 ships with npm v10. A user would need to actively downgrade to an older version of npm. So I'm not sure who would benefit from allowing older versions.

Even the latest Node v18 ships with npm v10: https://nodejs.org/en/about/previous-releases

matz3: Alright, then let's go with `>= 10`

lib/processors/jsdoc/jsdocGenerator.js
d3xter666: Thanks!

matz3 (approved):

PR #994, opened 2024-02-27
[INTERNAL] tests: remove packages 'chai', 'chai-fs' and 'recursive-dir'

flovogt:
Get rid of `chai`, `chai-fs` and `recursive-dir`. 

`chai-fs` is not compatible with `chai@v5`. Therefore upgrade to latest `chai` is not possible.
`recursive-dir` is no longer required because node supports option `recursive: true` for native `readdir`


RandomByte (approved):
lgtm

PR #1001, opened 2024-03-18
[FIX] Fix test failures

d3xter666:


matz3:
This is not relevant for consumers, or is it? I would therefore not expect a [FIX] prefix but rather [INTERNAL].

PR #1014, opened 2024-05-10
[FIX] bundle/Builder: Correct bundling of resources with empty source map

RandomByte:
If a resource features an empty source map (which is not the same as
having no source map), the bundle's source map generation needs to
handle this differently from cases where mappings exist.

It must not prefix the empty mapping with a string "AAAA,"
since the danglíng comma will result in a corrupt source map.

However, it should still add the mapping for the first line/column
"AAAA". Otherwise the source map of the preceding module in the bundle
would be applied.

Resolves the root cause of https://github.com/SAP/openui5/issues/4035
DINC0139775


RandomByte:
Also see my comment at https://github.com/SAP/openui5/issues/4035#issuecomment-2104570168

d3xter666 (approved):
LGTM

matz3 (approved):

RandomByte:
Force pushed to re-trigger PR checks that did not start

PR #1018, opened 2024-05-15
[INTERNAL] generateVersionInfo: Only process dependencies of type 'library'

RandomByte:
Ignore any other project types, even if the project contains a .library file.

Currently, projects of type 'module' will return null as their
namespace, which is then incorrectly used as a string in the constructed
glob pattern.

Projects of type 'library' always provide a valid namespace.

Relates to https://github.com/SAP/ui5-server/pull/656


d3xter666 (approved):
LGTM

PR #1032, opened 2024-06-18
[INTERNAL] ES2023 Support

flovogt:
Ensures valid build outputs also for projects using latest ES2023 syntax.

Hashbangs e.g. `#!/usr/bin/env node` are allowed in source files but won't be included in bundles (e.g. Component-preload.js) by UI5 Tooling.

JIRA: CPOUI5FOUNDATION-850

cherry-pick of 30301b24f27beb220c73e0bcc255f1ec806ab819


lib/lbt/bundle/Builder.js
```diff
@@ -36,6 +36,9 @@ function makeStringLiteral(str) {
 		return strReplacements[char];
 	}) + "'";
 }
+function removeHashbangs(str) {
+	return str.replace(/^#!(.*)/g, "");
```
RandomByte, line 40:
```suggestion
	return str.replace(/^#!(.*)/, "");
```

Global flag should not be necessary

RandomByte (changes requested):

lib/lbt/bundle/Builder.js
flovogt: done

RandomByte (approved):

PR #1036, opened 2024-06-24
[BREAKING] Discontinue bundling of JavaScript modules as string

flovogt:
The UI5 bundler packages JavaScript files that are identified as "requiresTopLevelScope" as a string, to be evaluated via "eval" at runtime. This behavior ensures that the script works as expected, e.g. with regards to implicit globals.
This "eval" runtime feature with be discontinued in UI5 2.0 because of security best practices (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval) and to comply with stricter CSP settings (unsafe-eval).
By using specVersion 4.0 an error is thrown when bundling a module as string and the module is not included in the bundle at all.

JIRA: CPOUI5FOUNDATION-794


test/lib/lbt/bundle/Builder.js
```diff
@@ -321,6 +330,58 @@ ${SOURCE_MAPPING_URL}=Component-preload.js.map
 			"myModuleUsingGlobalScope.js",
```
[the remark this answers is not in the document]
codeworrior: @flovogt: you make an even bigger point here :-( 

First of all, even if it is hard to update the aggregation, I guess it would be doable. But there's a bigger problem with sections of mode `bundleInfo`. The filters for such sections are redundant to the real bundle definition. And, when building the `bundleInfo` section, we don't look into the content but just resolve the filters based on the names. If a module matches the filters but isn't eligible for preload-bundling (e.g. uses globals), then we would still add it to the bundleInfo, although it's not contained there.

Constant solution would be to analyse modules in all cases, but that for sure has a performance penalty (if we don't add caching for that information).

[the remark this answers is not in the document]
flovogt: The team agreed on removing the modules / submodules information also form the bundle info in case the file content is not part of the bundle

lib/lbt/bundle/Builder.js
```diff
@@ -542,15 +551,36 @@ class BundleBuilder {
...
+	async writeBundleInfos(sections) {
 		this.outW.ensureNewLine();
 
 		let bundleInfoStr = "";
 		if ( sections.length > 0 ) {
 			bundleInfoStr = "sap.ui.loader.config({bundlesUI5:{\n";
-			sections.forEach((section, idx) => {
-				if ( idx > 0 ) {
+			let initial = true;
+			for (let idx=0; idx < sections.length; idx++) {
```
RandomByte, line 574:
```suggestion
			for (let idx = 0; idx < sections.length; idx++) {
```

RandomByte (changes requested):
I'm missing some tests for the case of specVersion 4 project with resources that require bundling as string

lib/lbt/bundle/Builder.js
```diff
@@ -542,15 +551,36 @@ class BundleBuilder {
...
+				const section = sections[idx];
+
+				let modules = await Promise.all(section.modules.map(this.checkForGlobals.bind(this)));
+				modules = modules.filter(($) => $) || [];
+
+				if (!initial) {
 					bundleInfoStr += ",\n";
+				} else {
+					initial=false;
```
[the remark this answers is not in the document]
flovogt: done

lib/lbt/bundle/Builder.js
flovogt: done

flovogt:
ui5 tooling integration tests are triggered: 534

RandomByte (dismissed):
LGTM

lib/lbt/bundle/Builder.js
```diff
@@ -542,15 +547,41 @@ class BundleBuilder {
 		});
 	}
 
-	writeBundleInfos(sections) {
+	logStringBundlingError(moduleName) {
+		log.error(
```
flovogt, line 551: @KlattG what you think. error message fine?

lib/lbt/bundle/Builder.js
```diff
@@ -542,15 +547,41 @@ class BundleBuilder {
...
+			"Module " + moduleName + " requires top level scope and can only be embedded as a string " +
+			"(requires 'eval'), which is not supported with specVersion 4.0 and newer");
```
KlattG, line 553:
```suggestion
			"(requires 'eval'), which is not supported with specVersion 4.0 and higher");
```

flovogt: done

lib/lbt/bundle/Builder.js
flovogt: done

KlattG (approved):
LGTM

lib/lbt/bundle/Builder.js
```diff
@@ -542,15 +547,41 @@ class BundleBuilder {
 		});
 	}
 
-	writeBundleInfos(sections) {
+	logStringBundlingError(moduleName) {
+		log.error(
+			"Module " + moduleName + " requires top level scope and can only be embedded as a string " +
+			"(requires 'eval'), which is not supported with specVersion 4.0 and higher");
```
matz3, line 553: Should we also give a hint to check our documentation to learn how to resolve this?
The warning message was always a bit confusing as it's not straight-forward to understand the scenario and reasoning behind, so I think it would help many users to get a link or hint to look into our documentation to learn what to do in this case (see https://github.com/SAP/ui5-tooling/pull/975)

flovogt: added the link to the future doccumenation

lib/lbt/bundle/Builder.js
```diff
@@ -542,15 +547,43 @@ class BundleBuilder {
 		});
 	}
 
-	writeBundleInfos(sections) {
+	logStringBundlingError(moduleName) {
+		log.error(
+			"Module " + moduleName + " requires top level scope and can only be embedded as a string " +
+			"(requires 'eval'), which is not supported with specVersion 4.0 and higher. " +
+			"Please checkout the UI5 Tooling documentation for further information " +
```
flovogt, line 554: @KlattG could you please check this error message again?

KlattG, line 554:
```suggestion
			"For more information, see the UI5 Tooling documentation " +
```

KlattG (changes requested):
Just one minor point, otherwise LGTM

lib/lbt/bundle/Builder.js
flovogt: done

lib/lbt/bundle/Builder.js
```diff
@@ -542,15 +547,43 @@ class BundleBuilder {
 		});
 	}
 
-	writeBundleInfos(sections) {
+	logStringBundlingError(moduleName) {
+		log.error(
+			"Module " + moduleName + " requires top level scope and can only be embedded as a string " +
+			"(requires 'eval'), which is not supported with specVersion 4.0 and higher. " +
+			"For more information, see the UI5 Tooling documentation  " +
```
matz3, line 554: Is there a purpose for the two spaces in between the text and the link?

KlattG, line 554:
```suggestion
			"For more information, see the UI5 Tooling documentation " +
```

flovogt: done

flovogt: good catch :-D

PR #1033, opened 2024-06-19
[FEATURE] Apply default values to bunde definitions for standard tasks

d3xter666:
JIRA: CPOUI5FOUNDATION-835

PR #1042, opened 2024-07-02
[FEATURE] Bundle `require` section with async flag for specVersion: 4.0

d3xter666:
JIRA: CPOUI5FOUNDATION-835

This change completes https://github.com/SAP/ui5-builder/pull/1033 with check for specVersion: 4.0 and enforcing default async flag for require section in `bundleDefinitions`


d3xter666:
The build fails as for some reason Azure tries to install Node 20.15.0, which has some isseues with Buffers.
With that particular version the issue is also reproducible locally

lib/processors/bundlers/moduleBundler.js
```diff
@@ -152,14 +152,12 @@ export default function({resources, options: {
 		ignoreMissingModules: false
 	}, bundleOptions);
 
+	// Take the project from the resources if available.
+	// It's needed for specVersion check. If unavailable,
+	// the specVersion would be treated as the most recent one.
+	const project = resources?.[0]?.getProject();
```
RandomByte, line 158: I'm afraid that `resources` might contain resources from multiple project, not only the one currently being built. Think for example of an app generating a self-contained bundle.

Similar to other cases, we should do the specVersion check in the individual tasks (in this case all the bundling tasks) and set the configuration there.

RandomByte (changes requested):

lib/processors/bundlers/moduleBundler.js
d3xter666: Ah, I see!

Thanks a lot! I'll try to cover this

d3xter666: Now the `taskUtil` is part of `moduleBundler`'s signature as we do that in other places

lib/processors/bundlers/moduleBundler.js
```diff
@@ -135,11 +135,12 @@ const log = getLogger("builder:processors:bundlers:moduleBundler");
  * @param {string} [parameters.options.targetUi5CoreVersion] Optional semver compliant sap.ui.core project version, e.g '2.0.0'.
 				This allows the bundler to make assumptions on available runtime APIs.
 				Omit if the ultimate UI5 version at runtime is unknown or can't be determined.
+ * @param {@ui5/project/build/helpers/TaskUtil|object} [parameters.taskUtil] TaskUtil
  * @returns {Promise<module:@ui5/builder/processors/bundlers/moduleBundler~ModuleBundlerResult[]>}
  * Promise resolving with module bundle resources
  */
 /* eslint-enable max-len */
-export default function({resources, options: {
+export default function({resources, taskUtil, options: {
```
RandomByte, line 143:
```suggestion
export default function({resources, options: {
```

Remove unused parameter

RandomByte (changes requested):

RandomByte (approved):

lib/processors/bundlers/moduleBundler.js
```diff
@@ -135,6 +135,7 @@ const log = getLogger("builder:processors:bundlers:moduleBundler");
  * @param {string} [parameters.options.targetUi5CoreVersion] Optional semver compliant sap.ui.core project version, e.g '2.0.0'.
 				This allows the bundler to make assumptions on available runtime APIs.
 				Omit if the ultimate UI5 version at runtime is unknown or can't be determined.
+ * @param {@ui5/project/build/helpers/TaskUtil|object} [parameters.taskUtil] TaskUtil
```
matz3, line 138: obsolete parameter? or was it already in use but undocumented?

flovogt: I have removed it in https://github.com/SAP/ui5-builder/pull/1036/files#diff-255002459f4f366a0dceae0db4927dd06d44894f887b3ba210884d9a399f48d1L138

lib/tasks/bundlers/utils/applyDefaultsToBundleDefinition.js
```diff
@@ -0,0 +1,31 @@
...
+			renderer: false,
+			sort: true,
+			declareRawModules: false,
+		};
+
+		// Since specVersion: 4.0 "require" section starts loading asynchronously.
+		// If specVersion cannot be determined, the latest spec is taken into account.
+		// This is a breaking change in specVersion: 4.0
+		if (section.mode === "require" && (!taskUtil || taskUtil.getProject().getSpecVersion().gte("4.0"))) {
+			defaultValues.async = true;
```
matz3, line 24: Is there a test that verifies this behavior?
The bundle tooling already defaults to "async=true", so I doubt that this line has any effect. My expectation would be that the flag has to be actively set to false in the else case, as this is the value that differs from the new default behavior.

I tested this with a self-contained build and even with specVersion 3.0 I get an async require section, which is not expected.

RandomByte: We should also document the "async" option in the respective JSDoc: https://github.com/SAP/ui5-builder/blob/main/lib/processors/bundlers/moduleBundler.js#L13

PR #1048, opened 2024-07-11
[INTERNAL] Async bundling for require section

d3xter666:
Require's section `async` flag has different default values based on the specVersion. However, async require is now build by default. 
We need to explicitly default `async=false` for specVersions lower than 4.0


matz3:
Please use [INTERNAL] as this fixes an issue of a feature that hasn't been released, so it is not relevant to be mentioned in the changelog.

matz3 (commented):
Works as expected 👍🏻

matz3:
Please also keep @RandomByte's comment in mind: https://github.com/SAP/ui5-builder/pull/1042#discussion_r1673712338

d3xter666:
@KlattG , would you take a look at this one, please?

lib/processors/bundlers/moduleBundler.js
```diff
@@ -35,8 +35,11 @@ const log = getLogger("builder:processors:bundlers:moduleBundler");
  *	<li>
  *		<code>require</code>: A 'require' section is transformed into a sequence of jQuery.sap.require calls. The
  *		list will be resolved like an include pattern list in any of the other sections and for each of the resolved
- *		modules, a jQuery.sap.require will be created. In case the ui5loader is available, 'sap.ui.requireSync' is
+ *		modules, a jQuery.sap.require will be created. In case the ui5loader is available, 'sap.ui.require' is
```
matz3, line 38: The usage of `jQuery.sap.require` has already been removed. This was the case in the past, but this text does not fit to the current codebase and should be updated.

lib/processors/bundlers/moduleBundler.js
```diff
@@ -35,8 +35,11 @@ const log = getLogger("builder:processors:bundlers:moduleBundler");
...
  *		used instead.
+ *		With specVersion: 4.0 this module comes with an `async` flag. When set to false, the modules
```
matz3, line 40: Is the moduleBundler aware of a specVersion? IMO it just has an API interface with certain defaults, but the processor is completely independent of any project configuration.

lib/processors/bundlers/moduleBundler.js
matz3: In addition: The `@typedef {object} ModuleBundleDefinitionSection` lists the available properties, and `async` is still missing there.

KlattG, line 38: Isn't this deprecated in UI5?
https://sapui5untested.int.sap.eu2.hana.ondemand.com/sapui5-sdk-internal/#/api/jQuery.sap%23methods/jQuery.sap.require

lib/processors/bundlers/moduleBundler.js
KlattG, line 40:
```suggestion
 *		With specVersion: 4.0, this module comes with an `async` flag. When set to false, the modules
```

lib/processors/bundlers/moduleBundler.js
```diff
@@ -35,8 +35,11 @@ const log = getLogger("builder:processors:bundlers:moduleBundler");
...
+ *		will be loaded using `sap.ui.requireSync` instead of `sap.ui.require`.
```
KlattG, line 41:
```suggestion
 *		are loaded using `sap.ui.requireSync` instead of `sap.ui.require`.
```

lib/processors/bundlers/moduleBundler.js
```diff
@@ -35,8 +35,11 @@ const log = getLogger("builder:processors:bundlers:moduleBundler");
...
+ *		**Note:** `sap.ui.requireSync` API is not available in UI5 version 2.x.
```
KlattG, line 42:
```suggestion
 *		**Note:** The `sap.ui.requireSync` API is not available in UI5 version 2.x.
```

KlattG (changes requested):

lib/processors/bundlers/moduleBundler.js
d3xter666: Done

lib/processors/bundlers/moduleBundler.js
d3xter666: Done

lib/processors/bundlers/moduleBundler.js
d3xter666: Done

lib/processors/bundlers/moduleBundler.js
d3xter666: Removed this line

lib/processors/bundlers/moduleBundler.js
```diff
@@ -33,10 +33,12 @@ const log = getLogger("builder:processors:bundlers:moduleBundler");
  *      the ui5loader is available.
  *	</li>
  *	<li>
- *		<code>require</code>: A 'require' section is transformed into a sequence of jQuery.sap.require calls. The
+ *		<code>require</code>: A 'require' section is transformed into a sequence of 'sap.ui.require' calls. The
  *		list will be resolved like an include pattern list in any of the other sections and for each of the resolved
- *		modules, a jQuery.sap.require will be created. In case the ui5loader is available, 'sap.ui.requireSync' is
- *		used instead.
+ *		modules, a 'sap.ui.require' will be created.
```
matz3, line 38: The description still does not fit to the actual behavior. Please check for example: https://github.com/SAP/ui5-builder/blob/7691b08c4cd8abc7fe3922ec913f80dca372ba4f/lib/lbt/bundle/Builder.js#L80-L85

There is only one sap.ui.require call per section, and the modules are loaded in parallel. Only when using `async: false`, they are loaded with individual `sap.ui.requireSync` calls in sequence.

d3xter666: What about now?
Is it better?

matz3: Yes, looks better 👍🏻

d3xter666:
The build fails as we need to publish specVersion: 4.0 in order tests to sun successfully

matz3:
> The build fails as we need to publish specVersion: 4.0 in order tests to sun successfully

Tests already fail in main as we're about to do the release. cc @flovogt 

Independent from that there are also two eslint errors introduced with this PR.

d3xter666:
Hi @KlattG,

I have updated the texts according to the comments above.
Would you take a look again at this PR, please?

lib/processors/bundlers/moduleBundler.js
```diff
@@ -75,6 +75,10 @@ const log = getLogger("builder:processors:bundlers:moduleBundler");
  * @property {boolean} [declareRawModules=false] Whether raw modules should be declared after jQuery.sap.global
  *		became available. With the usage of the ui5loader, this flag should be set to 'false'
  * @property {boolean} [sort=true] Whether the modules should be sorted by their dependencies
+ * @property {boolean} [async=true] Whether the require section of the module should be asynchronous.
```
KlattG, line 78:
```suggestion
 * @property {boolean} [async=true] Whether the `require` section of the module should be loaded asynchronously.
```

lib/processors/bundlers/moduleBundler.js
```diff
@@ -75,6 +75,10 @@ const log = getLogger("builder:processors:bundlers:moduleBundler");
...
+ * 		When set to true, the modules will be loaded using sap.ui.require instead of sap.ui.requireSync.
```
KlattG, line 79:
```suggestion
 * 		When set to true, the modules are loaded using `sap.ui.require` instead of `sap.ui.requireSync`.
```

lib/processors/bundlers/moduleBundler.js
```diff
@@ -75,6 +75,10 @@ const log = getLogger("builder:processors:bundlers:moduleBundler");
...
+ * 		The latter API is not available in UI5 version 2.x.
+ * 		Note: This property is available only for mode=require.
```
KlattG, line 81:
```suggestion
 * 		Note: This property is available only for `mode=require`.
```

KlattG (changes requested):

lib/processors/bundlers/moduleBundler.js
```diff
@@ -75,6 +75,10 @@ const log = getLogger("builder:processors:bundlers:moduleBundler");
  * @property {boolean} [declareRawModules=false] Whether raw modules should be declared after jQuery.sap.global
  *		became available. With the usage of the ui5loader, this flag should be set to 'false'
  * @property {boolean} [sort=true] Whether the modules should be sorted by their dependencies
+ * @property {boolean} [async=true] Whether the `require` section of the module should be loaded asynchronously.
+ * 		When set to true, the modules are loaded using `sap.ui.require` instead of `sap.ui.requireSync`.
+ * 		The latter API is not available in UI5 version 2.x.
+ * 		Note: This property is available only for `mode=require`.
```
KlattG, line 81:
```suggestion
 * 		**Note:** This property is available only for `mode=require`.
```

KlattG (commented):
Just one small suggestion, otherwise fine.

lib/processors/bundlers/moduleBundler.js
matz3: Thanks!

lib/processors/bundlers/moduleBundler.js
```diff
@@ -75,6 +75,10 @@ const log = getLogger("builder:processors:bundlers:moduleBundler");
  * @property {boolean} [declareRawModules=false] Whether raw modules should be declared after jQuery.sap.global
  *		became available. With the usage of the ui5loader, this flag should be set to 'false'
  * @property {boolean} [sort=true] Whether the modules should be sorted by their dependencies
+ * @property {boolean} [async=true] Whether the `require` section of the module should be loaded asynchronously.
+ * 		When set to true, the modules are loaded using `sap.ui.require` instead of `sap.ui.requireSync`.
```
codeworrior, line 79: Should we mention that each section creates one sap.ui.require call (or alternatively, that the modules are required in parallel)?

matz3:
@KlattG could you please check my changes in https://github.com/SAP/ui5-builder/pull/1048/commits/3a8289bb4fdc151e937af8478f19b2c1af39a9ed?

KlattG (approved):
LGTM

matz3 (approved):

PR #1065, opened 2024-07-22
[INTERNAL] lib/processors/jsdoc: support optional methods also in classes

codeworrior:
So far, the documentation build only allowed optional methods in interfaces. But UI5 in several places specifies methods on a base class level without implementing them. Subclasses can then implement such a method to get a certain behavior (example: Control#getAccessibilityInfo).

TypeScript (meanwhile?) allows to describe this as optional methods in the base class.

To allow to use this and thereby to better describe those methods in the TypeScript types, the JSDoc template lifts its limitation and writes out the 'optional' marker also for methods in classes.

Note: this is now even possible for static class methods, although this is not used yet in the framework.

Cherry-picked from SAP/openui5@a7ccde144.


RandomByte (approved):
Test failures are unrelated to this PR

PR #1075, opened 2024-07-26
[FIX] manifestEnhancer: Fix fallbackLocale handling / Improve error handling

matz3:


lib/processors/manifestEnhancer.js
```diff
@@ -368,6 +368,11 @@ class ManifestEnhancer {
 			return;
 		}
 
+		if (!this.manifest["sap.app"]?.id) {
```
codeworrior, line 371: Why the check for `id` when the processing requires the `i18n` property?

matz3: The ID is essential for determining the namespace of the project. It will be used to check whether a bundle is located within the project.

On the other hand sap.app/i18n is not required to be set. For libraries it is not relevant at all (as there is sap.ui5/library/i18n) and for applications it does not need to be defined when the default bundle location at `./i18n/i18n.properites` is used.

RandomByte (approved):

PR #1079, opened 2024-08-28
[INTERNAL] Migrate project to TypeScript

RandomByte:


flovogt:
The team did a proof of concept to get an overview of how much effort it would be to migrate the UI5 Tooling repositories from JavaScript to TypeScript. We came far.
However, we have identified some challenges that made us decide against a migration for now.

## Code
- Workerpool can't transpile TypeScript on the fly during test execution
- Runtime type checks are redundant when using TypeScript; however, they are still needed for JavaScript consumers

## Documentation
- There is no mature concept to produce cross-repo public API documentation. [TypeDoc](https://typedoc.org/) might bridge this gap in the future.

## Types
- How should we ship type definitions for UI5 Tooling packages for internal (UI5 Tooling) development as well as external consumption (e.g. custom middleware)? External consumers might need specVersion-dependent types.

## Test Setup
- Existing tests mock a lot of objects and violate TypeScript types all the time

PR #1080, opened 2024-09-05
[FIX] manifestEnhancer: Only use valid files for supportedLocales

matz3:
This fixes two problems that could have occurred:

A properties file with an invalid locale was still taken into the list of supported locales, which then caused a runtime exception in the ResourceBundle as it validates the input.

Another problem was that properties files could have a valid name according to BCP47, but the file won't be ever requested with that name.
This is due to the fact that the ResourceBundle does use the legacy Java locale format (using underscores instead of dashes) for the request URL.

In both cases, the properties file is now ignored and no entry for the supportedLocales is created.

Only locales that are valid according to the legacy Java locale format are considered.
However, there is one special case: sr_Latn is also requested by the UI5 runtime, although it contains a BCP47 script, which is not valid according to the legacy Java locale format.

### DIFF
--- a/lib/processors/manifestEnhancer.js
+++ b/lib/processors/manifestEnhancer.js
@@ -231,7 +231,7 @@ class ManifestEnhancer {
 				if (bcp47Locale) {
 					supportedLocales.push(bcp47Locale);
 				} else {
-					log.warn(`Skipping invalid file '${fileName}' for bundle '${i18nBundleUrl}'`);
+					log.warn(`Ignoring unexpected locale in filename '${fileName}' for bundle '${i18nBundleUrl}'`);
 				}
 			}
 		});
--- a/lib/processors/manifestEnhancer.js
+++ b/lib/processors/manifestEnhancer.js
@@ -11,11 +11,14 @@ const APP_DESCRIPTOR_V22 = new Version("1.21.0");
  * Matches a legacy Java locale string, which is the format used by the UI5 Runtime (ResourceBundle)
  * to load i18n properties files.
  * Special case: "sr_Latn" is also supported, although the BCP47 script part is not supported by the Java locale format.
+ *
+ * Variants are limited to the format from BCP47, but with underscores instead of hyphens.
  */
-//                          [     language     ]    [    region    ]    [  variants  ]
-const rLegacyJavaLocale = /^([a-z]{2,3}|sr_Latn)(?:_([A-Z]{2}|\d{3})(?:_([a-zA-Z0-9]+))?)?$/;
+//                          [     language     ]    [    region    ][                variants               ]
+const rLegacyJavaLocale = /^([a-z]{2,3}|sr_Latn)(?:_([A-Z]{2}|\d{3})((?:_[0-9a-zA-Z]{5,8}|_[0-9][0-9a-zA-Z]{3})*)?)?$/;
 
-const sapSupportabilityLocales = ["saptrc", "sappsd", "saprigi"];
+// See https://github.com/SAP/openui5/blob/a8f36e430f1fac172eb705811da4a2af25483408/src/sap.ui.core/src/sap/base/i18n/ResourceBundle.js#L76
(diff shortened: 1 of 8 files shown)


lib/processors/manifestEnhancer.js
```diff
@@ -166,7 +204,11 @@ class ManifestEnhancer {
 				supportedLocales.push("");
 			} else if (fileNameWithoutExtension.startsWith(i18nBundlePrefix)) {
 				const locale = fileNameWithoutExtension.replace(i18nBundlePrefix, "");
-				supportedLocales.push(locale);
+				if (isValidPropertiesFilenameLocale(locale)) {
+					supportedLocales.push(locale);
+				} else {
+					log.verbose(`Skipping invalid locale '${locale}' for bundle '${i18nBundleUrl}'`);
```
matz3, line 210: Still open: What should happen in this case? Maybe rather an warning or even error log?

matz3: Decision: Log a warning and explain in commit message that the warning can be suppressed by manually maintaining the "supportedLocales".

matz3: Warning added via https://github.com/SAP/ui5-builder/pull/1080/commits/11ab19f778cfe4a9cae6cd31efadc6b4e797ef6c

> To be checked: Built-in handling of private extensions at runtime (1Q, 2Q, 3Q).

This just turned out to be a confusion based on a wrong comment / commit message. It referred to `variants`, but actually it was supposed to be `extensions`.

Extensions are not considered by the runtime code at all (i.e. can't appear in requests).
Private use is checked for the special SAP supportability locales, but the resulting request does not include an extension or private use section but will rather be e.g. `en_US_saptrc` (where `saptrc` is a variant).

lib/processors/manifestEnhancer.js
```diff
@@ -165,8 +226,13 @@ class ManifestEnhancer {
...
 				supportedLocales.push("");
 			} else if (fileNameWithoutExtension.startsWith(i18nBundlePrefix)) {
-				const locale = fileNameWithoutExtension.replace(i18nBundlePrefix, "");
-				supportedLocales.push(locale);
+				const fileNameLocale = fileNameWithoutExtension.replace(i18nBundlePrefix, "");
+				const bcp47Locale = getBCP47LocaleFromPropertiesFilename(fileNameLocale);
+				if (bcp47Locale) {
+					supportedLocales.push(bcp47Locale);
+				} else {
+					log.warn(`Skipping invalid file '${fileName}' for bundle '${i18nBundleUrl}'`);
```
RandomByte, line 234:
```suggestion
					log.warn(`Ignoring unexpected locale in filename '${fileName}' for bundle '${i18nBundleUrl}'`);
```

RandomByte (changes requested):

codeworrior (commented):
I'm fine with the code change. I struggled a bit wit hthe commit message. Would you mind detailing what *"This is due to the fact that the ResourceBundle does use the legacy Java locale format for the request URL."* means?

I guess you refer to the fact that not very possible part of a BCP47 language tag will be used by ResourceBundle for the outgoing requests (e.g. the script is currently ignored except for sr_Latn).

The references in the runtime code are meanwhile outdated (code moved to LanguageFallback), but they use a commit ID, so they continue to work.

codeworrior:
I didn't doubt that the comment is in general right (it is). It just took me a while to get what exactly this means.

Adding some details or examples would have helped to understand the problem better. The difference between "_" and "-" is one such detail. The runtime also won't sent out certain subtags, e.g. the script (except for sr_Latn) and others (e.g. private extensions other than those for the pseudo languages).

Well, but I don't want to overcomplicate it.

matz3:
@codeworrior thanks for your feedback. I've updated the commit message and also updated the code references.

codeworrior (approved):
Thank you for the updates! LGTM.

PR #1102, opened 2025-01-10
[FIX] minify: In case of error, include full resource path in the error messsage

RandomByte:
As suggested by Vest in https://github.com/SAP/ui5-tooling/issues/890#issuecomment-2580779768, this should make it easier to identify the resource causing an error in the minify task.

### DIFF
--- a/lib/processors/minifier.js
+++ b/lib/processors/minifier.js
@@ -274,6 +274,7 @@ export default async function({
 			filename,
 			dbgFilename,
 			code,
+			resourcePath,
 			sourceMapOptions
 		}, taskUtil);
 		resource.setString(result.code);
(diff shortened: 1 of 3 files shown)


lib/processors/minifier.js
```diff
@@ -274,6 +274,7 @@ export default async function({
 			filename,
 			dbgFilename,
 			code,
+			resourcePath: resource.getPath(),
```
matz3, line 277: `resourcePath` is already available as local variable. Or is the path changed in between?

RandomByte: Thanks, you're right! Corrected 👍

matz3 (approved):

PR #1109, opened 2025-03-26
[FEATURE]: Rewrite dataSource uri for sap.cloud applications

schiwekM:
Hi colleagues,

currently the package '@sap/ui5-builder-webide-extension' rewrites the dataSource uri and removes any leading slash for sap.cloud applications. 
This is needed, because only when no leading slash exists will the source be correctly queried via its destination, which is defined in sap.cloud.

Because the package didn't receive any update in 3 years, and their ui5 tooling dependencies are outdated, I open this PR to add the rewrite to the native UI5 Tooling to replace the '@sap/ui5-builder-webide-extension' package.

BR,
Marten

### DIFF
--- a/lib/processors/manifestEnhancer.js
+++ b/lib/processors/manifestEnhancer.js
@@ -258,6 +258,26 @@ class ManifestEnhancer {
 		}
 	}
 
+	processDataSourceUri() {
+		const sapCloud = this.manifest["sap.cloud"];
+		const sapApp = this.manifest["sap.app"];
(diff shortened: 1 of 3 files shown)


schiwekM:
Closed due to https://github.com/ui5-community/ui5-ecosystem-showcase/blob/main/packages/ui5-task-zipper/lib/zipper.js#L32. Only Fiori tools need to be updated to use the task zipper for it. Thanks for the hint Florian.

PR #1132, opened 2025-07-14
[FIX] bundle/Builder: Skip source map for empty or trivia-only files

RandomByte:
Follow-up of https://github.com/SAP/ui5-builder/pull/1014

Providing an "AAAA" mapping for empty files or files containing only
comments produces a warning in the Safari browser, which then ignores
the whole source map:

Screenshot 2025-07-14 at 11 46 19

To resolve this, do not create any mapping for such files. In my
testing, the behavior in Chrome appears to be the same with and without
the mapping.


d3xter666 (approved):
LGTM

PR #1135, opened 2025-07-30
[FIX] Resource filtering: Add filtering of resources tagged on previous tasks

cotchere:

## Pull Request Checklist
- [x] Reviewed the [Contributing Guidelines](https://github.com/SAP/ui5-tooling/blob/main/CONTRIBUTING.md#-contributing-code)
    + Especially the [How to Contribute](https://github.com/SAP/ui5-tooling/blob/main/CONTRIBUTING.md#how-to-contribute) section 
- [x] [No merge commits](https://github.com/SAP/ui5-tooling/blob/main/docs/Guidelines.md#no-merge-commits)
- [x] [Correct commit message style](https://github.com/SAP/ui5-tooling/blob/main/docs/Guidelines.md#commit-message-style)

### DIFF
--- a/lib/tasks/minify.js
+++ b/lib/tasks/minify.js
@@ -28,7 +28,16 @@ import fsInterface from "@ui5/fs/fsInterface";
 export default async function({
 	workspace, taskUtil, options: {pattern, omitSourceMapResources = false, useInputSourceMaps = true
 	}}) {
-	const resources = await workspace.byGlob(pattern);
+	let filteredWorkspace = workspace;
+	if (taskUtil) {
+		filteredWorkspace = taskUtil.resourceFactory.createFilterReader({
+			reader: workspace,
+			callback: function(resource) {
+				return !taskUtil.getTag(resource, taskUtil.STANDARD_TAGS.OmitFromBuildResult);
+			}
+		});
+	}
+	const resources = await filteredWorkspace.byGlob(pattern);
 	const processedResources = await minifier({
 		resources,
 		fs: fsInterface(workspace),
--- a/test/lib/tasks/minify.integration.js
+++ b/test/lib/tasks/minify.integration.js
@@ -3,6 +3,7 @@ import sinonGlobal from "sinon";
 import minify from "../../../lib/tasks/minify.js";
 import * as resourceFactory from "@ui5/fs/resourceFactory";
 import DuplexCollection from "@ui5/fs/DuplexCollection";
+import {createTaskUtil} from "../../utils/taskUtilHelper.js";
 
 // Node.js itself tries to parse sourceMappingURLs in all JavaScript files. This is unwanted and might even lead to
 // obscure errors when dynamically generating Data-URI soruceMappingURL values.
@@ -28,16 +29,16 @@ test.afterEach.always((t) => {
 	t.context.sinon.restore();
 });
 
+
 test.serial("integration: minify omitSourceMapResources=true", async (t) => {
-	const taskUtil = {
+	const taskUtil = createTaskUtil(t, {
 		setTag: t.context.sinon.stub(),
 		STANDARD_TAGS: {
 			HasDebugVariant: "1️⃣",
 			IsDebugVariant: "2️⃣",
 			OmitFromBuildResult: "3️⃣"
 		},
-		registerCleanupTask: t.context.sinon.stub()
-	};
+	});
 	const {reader, writer, workspace} = createWorkspace();
 	const content = `
 function test(paramA) {
@@ -107,15 +108,14 @@ test();`;
 });
 
 test.serial("integration: minify omitSourceMapResources=false", async (t) => {
-	const taskUtil = {
+	const taskUtil = createTaskUtil(t, {
 		setTag: t.context.sinon.stub(),
 		STANDARD_TAGS: {
 			HasDebugVariant: "1️⃣",
 			IsDebugVariant: "2️⃣",
 			OmitFromBuildResult: "3️⃣"
-		},
-		registerCleanupTask: t.context.sinon.stub()
-	};
+		}
+	});
 	const {reader, writer, workspace} = createWorkspace();
 	const content = `
 function test(paramA) {
@@ -276,15 +276,14 @@ ${SOURCE_MAPPING_URL}=test.js.map`;
 });
 
 test.serial("integration: minify error", async (t) => {
-	const taskUtil = {
+	const taskUtil = createTaskUtil(t, {
 		setTag: t.context.sinon.stub(),
 		STANDARD_TAGS: {
 			HasDebugVariant: "1️⃣",
 			IsDebugVariant: "2️⃣",
 			OmitFromBuildResult: "3️⃣"
-		},
-		registerCleanupTask: t.context.sinon.stub()
-	};
+		}
+	});
 	const {reader, workspace} = createWorkspace();
 	const content = `
 // Top level return will cause a parsing error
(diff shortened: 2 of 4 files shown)


cotchere:
Resources that have been tagged in previous tasks, such as with OmitFromBuildResult, are currently not taken into account by the minify task. As a result, these resources are still written to the workspace, even though they should be excluded from the build output. To ensure correct build results, the minify task must check resource tags and filter out any resources that have been marked (e.g., OmitFromBuildResult), preventing them from being written to the workspace.

RandomByte:
Thank you very much for the contribution. I just need some help in understanding the problem you are attempting to fix.

Do I get it right, [ui5-task-adaptation](https://github.com/SAP/ui5-task-adaptation) is setting the `OmitFromBuildResult` tag for a couple of resources, however they still appear in the final build result (i.e. `dist`-directory)?

If that is the case, I think we would need a different fix. The ProjectBuilder is supposed to filter all resources before writing them to disk, skipping any that have the `OmitFromBuildResult` tag: 
https://github.com/SAP/ui5-project/blob/acc032c6061ae1d80e1ee7f7ff8660900388c8a0/lib/build/ProjectBuilder.js#L400-L405

Could you please describe the current and the desired build output, maybe with an example we can reproduce? Do you see the actual files being unexpectedly written to the build result or do you see them in other resources like bundles?

Note that the `OmitFromBuildResult` tag is supposed to only affect the build **output**. It is expected that resources with that tag are still processed by most tasks. Also the server still provides them.

cotchere:
Hi @RandomByte,

I'm currently trying to resolve the following issue:

When a task executed before the minify task, such as [ui5-task-adaptation](https://github.com/SAP/ui5-task-adaptation), marks a resource with the OmitFromBuildResult tag, the minified versions of that resource are still being generated and written to the build output. This results in unintended artifacts being included in the final build.

Example:
Before Build:
```
webapp/
└ changes/
    └ coding/
        └─ example-script.js
```
If example-script.js is tagged with OmitFromBuildResult by ui5-task-adaptation, the build output currently looks like this:

After Build:

dist/
└ changes/
    └ coding/
        ├ example-script-dbg.js
        ├ example-script.js.map

Although the original file (example-script.js) is correctly omitted, the minify task still processes it and writes derivative files (-dbg.js, .map) to the output. The minify task does currently not filter resources by tags, but processes all resources in the workspace. Ideally, any resource marked with OmitFromBuildResult should be completely excluded from the build output, including any minified or debug variants.

RandomByte:
Thanks, I understand better now.

I think in order to stick with the design of the `OmitFromBuildResult` tag, a "better" fix would be for the minify task to carry over the tags from the input resource onto the derived resources. Do you think that would resolve your issue as well?

RandomByte:
Thanks. I agree with your points, it's certainly not ideal in this case.

We can also discuss introducing a new tag for this purpose. I understand you want those resources to be ignored by any subsequent tasks? We could add a tag for that. But then we should probably also have a mechanism in place that automatically filters resources for that tag for all tasks. Otherwise I think it's too likely for (custom-) tasks to forget to implement that filtering.

But even with this new tag, I think we would still need to make the change I proposed earlier: the minify task must carry over such tags to the derived resources. So maybe it's reasonable to make that change first and eventually optimize things later? What do you think? Did I miss something?

lib/tasks/minify.js
```diff
@@ -44,17 +44,28 @@ export default async function({
 		resource, dbgResource, sourceMapResource, dbgSourceMapResource
 	}) => {
 		if (taskUtil) {
-			taskUtil.setTag(resource, taskUtil.STANDARD_TAGS.HasDebugVariant);
```
RandomByte, line 47: Thank you for making the changes.

I think this whole block got a little complicated now.
Also, if a resource is tagged as omitted, and if there is a debug source map resource, that resource is no longer tagged as `IsDebugVariant`. This tag is relevant for later tasks though.

How about leaving the existing code as-is and just adding another block above? Maybe something like this:

```js
		if (taskUtil) {
			// --- New if clause:
			if (taskUtil.getTag(resource, taskUtil.STANDARD_TAGS.OmitFromBuildResult)) {
				// Carry over OmitFromBuildResult from input resource to all derived resources
				taskUtil.setTag(dbgResource, taskUtil.STANDARD_TAGS.OmitFromBuildResult);
				taskUtil.setTag(sourceMapResource, taskUtil.STANDARD_TAGS.OmitFromBuildResult);
			}
			// --- Existing code from here:
			taskUtil.setTag(resource, taskUtil.STANDARD_TAGS.HasDebugVariant);
			taskUtil.setTag(dbgResource, taskUtil.STANDARD_TAGS.IsDebugVariant);
			taskUtil.setTag(sourceMapResource, taskUtil.STANDARD_TAGS.HasDebugVariant);
			if (omitSourceMapResources) {
				taskUtil.setTag(sourceMapResource, taskUtil.STANDARD_TAGS.OmitFromBuildResult);
			}
			if (dbgSourceMapResource) {
				taskUtil.setTag(dbgSourceMapResource, taskUtil.STANDARD_TAGS.IsDebugVariant);
				if (omitSourceMapResources) {
					taskUtil.setTag(dbgSourceMapResource, taskUtil.STANDARD_TAGS.OmitFromBuildResult);
				}
			}
		}
```

Note that it's also not necessary to tag the input resource again.

test/lib/tasks/minify.js
```diff
@@ -1,22 +1,24 @@
...
+import {createTaskUtil} from "../../utils/taskUtilHelper.js";
 
 test.beforeEach(async (t) => {
 	const sinon = t.context.sinon = sinonGlobal.createSandbox();
 	t.context.workspace = {
 		byGlob: sinon.stub().resolves(["resource A", "resource B"]),
 		write: sinon.stub().resolves()
 	};
-	t.context.taskUtil = {
+	t.context.taskUtil = createTaskUtil(t, {
```
RandomByte, line 12: I'm afraid these changes can be reverted, right? `createFilterReader` is no longer used

test/lib/tasks/minify.integration.js
```diff
@@ -341,3 +340,67 @@ return;`;
...
+	});
+	const testResource2 = resourceFactory.createResource({
+		path: testFilePath2,
+		string: testFileContent2
+	});
+
+	await reader.write(testResource1);
+	await reader.write(testResource2);
+
+	const taskUtil = createTaskUtil(t);
```
RandomByte, line 364: Let's rather create a mock like before and have the getTag() method always return `true` for this test. Something like this should hopefully work:

```js
{
	setTag: t.context.sinon.stub(),
	getTag: t.context.sinon.stub().returns(true),
	STANDARD_TAGS: {
		HasDebugVariant: "1️⃣",
		IsDebugVariant: "2️⃣",
		OmitFromBuildResult: "3️⃣"
	},
	registerCleanupTask: t.context.sinon.stub()
}
```

Then we can remove the `taskUtilHelper`.

My reasoning here is that `taskUtilHelper` is basically re-implementing the functionality of the real TaskUtil. I would like to avoid that in tests. IMHO we should test with _either_ use original _or_ a "static" mock of it.

RandomByte (changes requested):

test/lib/tasks/minify.integration.js
cotchere: done

test/lib/tasks/minify.integration.js
```diff
@@ -341,3 +345,81 @@ return;`;
...
+		let res = await writer.byPath(filePath);
+		if (!res) {
+			t.fail(`Did not find ${filePath} in target locator although it should be present`);
+		}
+		const fileName = filePath.split("/").pop();
+		const expected = `function test(t){var o=t;console.log(o)}test();\n${SOURCE_MAPPING_URL}=${fileName}.map`;
+		const content = await res.getString();
+		t.deepEqual(content, expected, "Correct file content");
+
+		let isOmitted = taskUtil.getTag(res, taskUtil.STANDARD_TAGS.OmitFromBuildResult);
```
RandomByte: I'm afraid calling the stubbed `getTag` method will always return the expected result, no matter whether minify actually called `setTag` correctly. We need to check whether the `setTag` stub got called correctly.

I will create a commit to update this test

RandomByte (approved):
Again, thank you for your contribution. I think this is good to go now. I will try to get some additional feedback in the team today and hope that we can merge the PR soon 👌

codeworrior (approved):
LGTM

RandomByte:
This has been released in UI5 CLI v4.0.23. Again, thank you very much for the contribution!

PR #1144, opened 2025-09-06
Update jsdoc template

codeworrior:


lib/processors/jsdoc/lib/ui5/template/publish.js
```diff
@@ -2451,6 +2551,18 @@ function validateAPIJSON(api) {
 	}
 
 	function checkClass(symbol) {
+		const prevTypeParams = currentTypeParameters;
+		if ( Object.hasOwn(symbol, "constructor") && symbol.constructor.typeParameters ) {
+			currentTypeParameters = Object.create(currentTypeParameters);
+			for (const typeParam of symbol.constructor.typeParameters) {
+				currentTypeParameters[typeParam.name] = typeParam;
+			}
+			// check type oarameters in a 2nd loop so that they can depend on each other
```
flovogt, line 2560:
```suggestion
			// check type parameters in a 2nd loop so that they can depend on each other
```

flovogt (changes requested):

lib/processors/jsdoc/lib/ui5/template/publish.js
codeworrior: I won't change the cherry picked commits. I'll rather fix it in openui5, then cherry pick the fix.

codeworrior: Done.

PR #1164, opened 2025-10-13
[FEATURE] Enhance generateFlexChangeBundle task to add new 'flexBundle' property to the manifest.json

d3xter666:
JIRA: CPOUI5FOUNDATION-1125


lib/tasks/bundlers/generateFlexChangesBundle.js
```diff
@@ -108,10 +117,10 @@ export default async function({workspace, taskUtil, options = {}}) {
...
 			return workspace.write(resource);
 		}));
-		// Add the sap.ui.fl dependency if a bundle has been created
-		if (processedResources.length > 0) {
-			await updateFLdependency();
-		}
+		// Add the sap.ui.fl dependency if a bundle has been created and flag if a bundle was created
+		bBundleCreated = processedResources.length > 0;
+		await updateManifestWithFlDependencyAndFlexBundleFlag(bBundleCreated);
+		manifestUpdated = true;
```
matz3: Calling `updateManifestWithFlDependencyAndFlexBundleFlag` here and having a `manifestUpdated` flag is not needed in case the function should be always called anyways, right?

It could just be called at the end of the function without a condition.

d3xter666: Done

lib/tasks/bundlers/generateFlexChangesBundle.js
```diff
@@ -108,10 +116,8 @@ export default async function({workspace, taskUtil, options = {}}) {
 			log.verbose("Writing flexibility changes bundle");
 			return workspace.write(resource);
 		}));
-		// Add the sap.ui.fl dependency if a bundle has been created
-		if (processedResources.length > 0) {
-			await updateFLdependency();
-		}
+		// Add the sap.ui.fl dependency if a bundle has been created and flag if a bundle was created
```
matz3: I feel like this comment doesn't relate to the code anymore. At least the added `and flag if a bundle was created` is not related to it.

d3xter666: Thanks!
It's actually partially valid- it will add the `sap.ui.fl` dependency in case the value is `true`.
I have updated it

matz3 (approved):

PR #1170, opened 2025-10-28
[INTERNAL] generateLibraryPreload: Add experimental bundle-info preload

RandomByte:
For internal PoC usage.

This change is not active in regular builds. To activate it, the environment variable `UI5_CLI_EXPERIMENTAL_BUNDLE_INFO_PRELOAD` needs to be set to any value.

JIRA: CPOUI5FOUNDATION-1165

### DIFF
--- a/lib/tasks/bundlers/generateLibraryPreload.js
+++ b/lib/tasks/bundlers/generateLibraryPreload.js
@@ -33,6 +33,32 @@ function getDefaultLibraryPreloadFilters(namespace, excludes) {
 	return filters;
 }
 
+function getExperimentalDefaultLibraryPreloadFilters(namespace, excludes) {
+	const filters = [
+		`${namespace}/library.js`,
+		`!${namespace}/**/*-preload.js`, // exclude all bundles
+		`!${namespace}/designtime/`,
+		`!${namespace}/**/*.designtime.js`,
+		`!${namespace}/**/*.support.js`
+	];
+
+	if (Array.isArray(excludes)) {
+		const allFilterExcludes = negateFilters(excludes);
+		// Add configured excludes at the end of filter list
+		allFilterExcludes.forEach((filterExclude) => {
+			// Allow all excludes (!) and limit re-includes (+) to the library namespace
+			if (filterExclude.startsWith("!") || filterExclude.startsWith(`+${namespace}/`)) {
+				filters.push(filterExclude);
+			} else {
+				log.warn(`Configured preload exclude contains invalid re-include: !${filterExclude.substr(1)}. ` +
+				`Re-includes must start with the library's namespace ${namespace}`);
+			}
+		});
+	}
+
+	return filters;
+}
+
 function getBundleDefinition(namespace, excludes) {
 	// Note: This configuration is only used when no bundle definition in ui5.yaml exists (see "skipBundles" parameter)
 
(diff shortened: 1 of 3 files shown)


RandomByte:
To be clarified: Should this be released as a pre-release of UI5 Builder and UI5 CLI (requires dedicated release pipeline) or as regular patches (with an empty changelog)? I would prefer the first option.

lib/tasks/bundlers/generateLibraryPreload.js
```diff
@@ -106,6 +106,46 @@ function getBundleDefinition(namespace, excludes) {
...
 
+function getBundleInfoPreloadDefinition(namespace, excludes) {
+	return {
+		name: `${namespace}/library-preload.js`,
+		sections: [{
+			mode: "preload",
+			filters: [
+				`${namespace}/library.js`,
+			],
+			resolve: true
```
codeworrior: As this now uses `resolve:true`, how is it ensured that only resources from the current lib are included (e.g. not `sap/ui/core/Lib` or `sap/ui/core/Core` which should be a dependency of all best practice library.js modules)? Is the workspaces only filled with those resources and that's the reason why ignoreMissingModules is set to true?

RandomByte:
> To be clarified: Should this be released as a pre-release of UI5 Builder and UI5 CLI (requires dedicated release pipeline) or as regular patches (with an empty changelog)? I would prefer the first option.

As discussed with @matz3, we tend to release this as a regular patch and consume it in the regular way.

matz3 (dismissed):

codeworrior:
Somehow review comments that I had made, went lost. Please wait with a merge, I'll redo the comments.

lib/tasks/bundlers/generateLibraryPreload.js
```diff
@@ -106,6 +107,73 @@ function getBundleDefinition(namespace, excludes) {
 	};
 }
 
+function getBundleInfoPreloadDefinition(namespace, excludes, coreVersion) {
+	const sections = [{
+		mode: "preload",
+		filters: [
```
codeworrior: IMO, this section also needs the default filters, esp. the excludes. If library.js depends on a module that should be excluded from standard preloads, then we should not include it in this section.

With the old configuration, this was not a issue as `resolve:true` was not used.

codeworrior (changes requested):

RandomByte:
> Somehow review comments that I had made, went lost. Please wait with a merge, I'll redo the comments.

Oh no! I force pushed, but PR comments should have been preserved. Did you maybe comment directly on the commit?

lib/tasks/bundlers/generateLibraryPreload.js
RandomByte: > Is the workspaces only filled with those resources and that's the reason why ignoreMissingModules is set to true?

Exactly.

The `generateLibraryPreload` task has no access to dependency resources. Therefore there is no need for excludes, but a need for `ignoreMissingModules` (which is now also active in Resolve.js)

codeworrior:
Not sure, maybe. Indeed, while I was reviewing, you force pushed in parallel.

lib/tasks/bundlers/generateLibraryPreload.js
RandomByte: Right, good point. Done

codeworrior (approved):
LGTM.

PR #1174, opened 2025-11-13
[INTERNAL] generateLibraryPreload: Rename experimental content bundle and add manifest.json flag

RandomByte:
* Rename the new bundle to _library-content.js
* Add a marker to the generated manifest.json: `"sap.ui5".library.bundleVersion` with a value `2`

JIRA: CPOUI5FOUNDATION-1165

### DIFF
--- a/lib/processors/manifestEnhancer.js
+++ b/lib/processors/manifestEnhancer.js
@@ -441,6 +441,12 @@ class ManifestEnhancer {
 
 		if (this.manifest["sap.app"].type === "library") {
 			await this.processSapUi5LibraryI18n();
+
+			if (process.env.UI5_CLI_EXPERIMENTAL_BUNDLE_INFO_PRELOAD) {
+				// Add flag indicating that the (currently experimental) bundleVersion 2 preload bundling is used
+				this.manifest["sap.ui5"].library ??= {};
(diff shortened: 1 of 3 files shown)


lib/processors/manifestEnhancer.js
```diff
@@ -441,6 +441,12 @@ class ManifestEnhancer {
 
 		if (this.manifest["sap.app"].type === "library") {
```
codeworrior, line 442: Isn't this too weak? If a developer forgets to set the manifest type correctly, we wouldn't generate the flag, even if the manifest is for a library (resides next to a library.js).

I guess that's rather a general question whether we want to validate parts of the manifest / complain about certain errors.

I would accept to not further check this and say the runtime error is a signal. But it might be hard for a developer to understand form the runtime error (no bundleVersion flag) that the real problem is the forgotten manifest type.

RandomByte: For applications it would be incorrect though, since their bundling method doesn't change...

Maybe we can introduce a more stringent manifest validation in UI5 CLI v5. Including comparing the manifest type to the project type

codeworrior: Sounds reasonable, yes.

matz3 (approved):

PR #1178, opened 2025-11-24
[FIX] Bundling: generateFlexChangesBundle should not rely on existing manifest.json

d3xter666:
When running the `generateFlexChangesBundle` task, `manifest.json` might not be present.

Follow-up of https://github.com/SAP/ui5-builder/pull/1164

### DIFF
--- a/test/lib/tasks/bundlers/generateFlexChangesBundle.js
+++ b/test/lib/tasks/bundlers/generateFlexChangesBundle.js
@@ -636,7 +636,7 @@ test("task does not fail when manifest.json is missing but changes exist", async
 		});
 	}, "Task should not fail when manifest.json is missing even with changes");
 
-	// Verify it's not trying to write to manifest.json
+	// Verify the task created the flexibility-bundle.json
 	const writeCalls = [];
 	for (let i = 0; i < placeholderWorkspace.write.callCount; i++) {
 		const call = placeholderWorkspace.write.getCall(i);
@@ -644,6 +644,9 @@ test("task does not fail when manifest.json is missing but changes exist", async
 		writeCalls.push(path);
 	}
 
+	// Should have written flexibility-bundle.json but NOT manifest.json
+	t.true(writeCalls.some((path) => path.includes("flexibility-bundle.json")),
+		"flexibility-bundle.json should be created");
 	t.false(writeCalls.some((path) => path.includes("manifest.json")),
 		"No manifest.json write should occur when manifest is missing");
 });
--- a/test/lib/tasks/bundlers/generateFlexChangesBundle.js
+++ b/test/lib/tasks/bundlers/generateFlexChangesBundle.js
@@ -579,3 +579,71 @@ test("flexBundle property overrides existing value when bundle is created", asyn
 	t.true(manifestContent["sap.ui5"].flexBundle, "flexBundle should be overridden to true when bundle is created");
 	t.deepEqual(manifestContent["sap.ui5"].dependencies.libs["sap.ui.fl"], {}, "sap.ui.fl dependency should be added");
 });
+
+test("task does not fail when manifest.json is missing and no changes exist", async (t) => {
+	const placeholderWorkspace = {
+		byGlob: async () => [], // No changes
+		byPath: async (path) => {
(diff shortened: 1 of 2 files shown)


test/lib/tasks/bundlers/generateFlexChangesBundle.js
```diff
@@ -579,3 +579,71 @@ test("flexBundle property overrides existing value when bundle is created", asyn
...
+				projectNamespace: "sap/ui/demo/app"
+			}
+		});
+	}, "Task should not fail when manifest.json is missing");
+
+	// No write calls should have been made since there's no manifest and no changes
+	t.is(placeholderWorkspace.write.callCount, 0, "workspace.write should not be called when manifest is missing");
+});
+
+test("task does not fail when manifest.json is missing but changes exist", async (t) => {
```
matz3, line 608: Is this aligned with the flex team? I can see that there was already an error with UI5 CLI v3, when you have a variant project with changes:

```
TypeError: Cannot read properties of null (reading 'getString')
    at readManifestMinUI5Version
```

Maybe there is actually some handling missing for `manifest.appdescr_variant` files?

d3xter666: Thanks!
I'll get in touch with them!

d3xter666: I did some digging into this, and here's what I found:

The adaptation project case (where we have `manifest.appdescr_variant` but no `manifest.json`) is actually a legit scenario. Test should be a valid one because:

**Why the manifest isn't updated:**
- `manifest.appdescr_variant` doesn't have a `sap.ui5` section at all 
- So there's nothing for the task to update even if it wanted to (no `sap.ui5.flexBundle` or `sap.ui5.dependencies` to set)

**But the task still does some work:**
- Collects all the flexibility changes from `/changes/*.{change,variant,ctrl_variant,...}`
- Bundles them into `flexibility-bundle.json` (or `changes-bundle.json` for older UI5 versions)

Before https://github.com/SAP/ui5-builder/pull/1164, the manifest update function was only called when a bundle was created (if (processedResources.length > 0)), so the missing manifest scenario never triggered.

With the flexBundle flag feature, I changed it to always call the update function (even when no bundle exists, to set flexBundle: false). This uncovered the edge case where manifest.json might not exist at all (like in adaptation projects). So I added the null check to make it robust.

The test ensures we handle this gracefully without throwing!

matz3: My point was the case where before #1164, the task already failed for adaptation projects (`manifest.appdescr_variant`). This made me wonder whether running the task is actually intended as it seems it never worked. I found internal projects that have added `--exclude-task generateFlexChangesBundle`. But an issue has never been reported to us.

My assumption is that this fix is correct, and projects that have previously excluded the task should enable it.

test/lib/tasks/bundlers/generateFlexChangesBundle.js
```diff
@@ -579,3 +579,71 @@ test("flexBundle property overrides existing value when bundle is created", asyn
...
+	}, "Task should not fail when manifest.json is missing even with changes");
+
+	// Verify it's not trying to write to manifest.json
+	const writeCalls = [];
+	for (let i = 0; i < placeholderWorkspace.write.callCount; i++) {
+		const call = placeholderWorkspace.write.getCall(i);
+		const path = call.args[0].getPath ? await call.args[0].getPath() : "unknown";
+		writeCalls.push(path);
+	}
+
```
matz3, line 646: I would expect an assertion here that verifies a bundle has been created. Or is the assumption that no bundles are created when no manifest.json exists?

d3xter666: Thanks! I've added an assertion to verify that `flexibility-bundle.json` is actually created when changes exist but no manifest is present: https://github.com/SAP/ui5-builder/pull/1178/commits/3cc83b7cf57f1196b3ff1df23ba56fbb17239b6d

matz3 (approved):

PR #1181, opened 2025-12-05
[INTERNAL] moduleBundler: Add JSDoc for depCache section

RandomByte:


lib/processors/bundlers/moduleBundler.js
```diff
@@ -45,12 +45,18 @@ const log = getLogger("builder:processors:bundlers:moduleBundler");
  *		the module itself.
  *		This requires the ui5loader to be available at build time and UI5 version 1.74.0 or higher at runtime.
  *	</li>
+ *	<li>
```
codeworrior, line 48: To get the full set of dependencies for XMLViews/-fragments it is, IIRC, essential that all used controls etc. can be found in dependencies.

codeworrior (approved):

PR #1196, opened 2026-02-16
[INTERNAL] Exclude knip checks from test script

d3xter666:


RandomByte (approved):
Also see https://github.com/SAP/ui5-cli/pull/831

PR #1202, opened 2026-03-23
Update jsdoc template

codeworrior:


lib/processors/jsdoc/lib/transformApiJson.js
```diff
@@ -987,22 +1012,69 @@ function transformer(sInputFile, sOutputFile, sLibraryFile, vDependencyAPIFiles,
...
+
+							if (subsectionFiles.length > 0) {
+								return { name: sectionName, hasSubsections: true, subsections: subsectionFiles };
+							}
+						} catch (error) {
+							log.error('Error scanning subsection directory:', path.join(componentDir, sectionName), error);
+						}
+					}
+					return sectionName;
+				});
```
d3xter666, line 1067: This implies that sections can have (eventually) only one level of depth.
If there are subfolders, they would be skipped.
Is this the requirement?

lib/processors/jsdoc/lib/transformApiJson.js
```diff
@@ -987,22 +1012,69 @@ function transformer(sInputFile, sOutputFile, sLibraryFile, vDependencyAPIFiles,
...
-	 * (FAQ data must be defined as *.md files in the <code>sFAQDir</code>)
+	 * Check for existence of sections data
+	 * (Sections data must be defined as *.md files in the <code>sSectionsDir</code>)
 	 * and add a boolean flag in case it exists
 	 *
 	 * @param oChainObject chain object
 	 */
-	function addFlagsForFAQData(oChainObject) {
-		if (!sFAQDir) {
+	function addFlagsForSectionsData(oChainObject) {
```
d3xter666, line 1021: Maybe the name is a bit missleading. My impression for this method is that it rather resolves the sections data / files

codeworrior: Hi Yavor, 

thanks for the review, but the changes are not ours, they're from another team and originate in the openui5 repo. So I won't change the code here, but rather keep it in sync with openui5/lib/jsdoc.

BR, Frank

lib/processors/jsdoc/lib/transformApiJson.js
codeworrior: Question goes to the SDK team.

d3xter666 (approved):

codeworrior:
Up-port to V5: https://github.com/UI5/cli/pull/1329

PR #1203, opened 2026-04-02
[FIX] Bundler: Improve error ESM handling

matz3:
- [FIX] Bundler: Reduce noise from ESM parse errors
- [FIX] Bundler: Skip ESM modules during bundling and log errors
- [FIX] Bundler: Also detect import.meta as ESM indicator

See commit messages for details.

### DIFF
--- a/lib/lbt/bundle/Builder.js
+++ b/lib/lbt/bundle/Builder.js
@@ -597,10 +597,11 @@ class BundleBuilder {
 		}
 	}
 
-	writeRequires(section) {
+	async writeRequires(section) {
 		if (section.modules.length === 0) {
 			return;
 		}
+
 		this.outW.ensureNewLine();
 		if (section.async === false) {
 			section.modules.forEach( (module) => {
@@ -767,7 +768,7 @@ async function rewriteDefine({moduleName, moduleContent, moduleSourceMap}) {
 	} catch (e) {
 		log.error(`Error while parsing ${moduleName}: ${e.message}`);
 		log.verbose(e.stack);
-		return {};
+		return null;
 	}
 
 	if ( ast.type === Syntax.Program &&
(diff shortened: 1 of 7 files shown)


d3xter666 (dismissed):

lib/lbt/resources/ResourcePool.js
```diff
@@ -62,8 +62,11 @@ async function determineDependencyInfo(resource, rawInfo, pool) {
 		try {
 			ast = parseJS(code, {comment: true});
 		} catch (err) {
-			log.error(`Failed to parse ${resource.name}: ${err.message}`);
+			log.verbose(`Failed to parse ${resource.name}: ${err.message}`);
 			log.verbose(err.stack);
+			if (err.message.includes("'import' and 'export' may appear only with 'sourceType: module'")) {
```
matz3: `Cannot use 'import.meta' outside a module` is another error that may appear

lib/lbt/bundle/Builder.js
```diff
@@ -597,17 +620,41 @@ class BundleBuilder {
...
+		// Filter out ESM modules that can't be loaded via sap.ui.require
+		const modules = [];
+		for (const module of section.modules) {
+			let resource = null;
+			try {
+				resource = await this.pool.findResourceWithInfo(module);
+			} catch {
+				// Module not found in pool - keep it in the list
+			}
+			if (resource?.info?.format === ModuleInfo.Format.ESM) {
```
RandomByte: I guess the filtering could also be done more broadly in https://github.com/SAP/ui5-builder/blob/4742cef7e3a2ffbe4408b6e03ee048c63083fb20/lib/lbt/bundle/Resolver.js#L124 or was it intentional to only filter for writing requires in particular?

I'm wondering if the lack of filtering in other parts of the bundle generation could lead to issues?

matz3: Good catch. I honestly didn't spend much time in looking for the best place to filter them out. I updated the code to filter them out much earlier and avoid building another array or filtering.

RandomByte (approved):

PR #1208, opened 2026-05-08
[FIX] lbt/bundle/Resolver: Ensure deterministic ordering of raw modules

matz3:
When multiple raw modules share a common dependency, the topological sort tie-breaks by input order. Previously, the input order depended on async resolution timing (pool insertion order), making the output non-deterministic.

Now modules are sorted by their filter definition order before topological sorting, ensuring that modules at the same dependency level appear in the order defined by the section's filter list.

### DIFF
--- a/lib/lbt/bundle/Resolver.js
+++ b/lib/lbt/bundle/Resolver.js
@@ -329,6 +329,15 @@ class BundleResolver {
 				return collectModulesForSection(section).
 					then( (modules) => {
 						if ( section.mode == SectionType.Raw && section.sort !== false ) {
+							// Sort modules by their filter definition order to ensure
+							// deterministic input for the topological sort.
+							// Topological sort preserves input order for tie-breaking,
+							// so this ensures modules at the same dependency level
+							// appear in the order defined by the section filters.
+							const filterOrder = new ResourceFilterList(section.filters, fileTypes);
+							modules.sort((a, b) => {
+								return filterOrder.matchIndex(a) - filterOrder.matchIndex(b);
+							});
 							// sort the modules in topological order
 							return topologicalSort(pool, modules).then( (modules) => {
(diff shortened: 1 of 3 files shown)


codeworrior:
As long as you don't assume any semantic behind this order (e.g. to compensate for not-declared dependencies), I'm fine with it. IMO, sorting the modules alphabetically would also result in a stable order.

matz3:
Well, I think there can be cases in which you want a specific order, and sorting alphabetically would break it:

Library1 provides 2 raw modules:
- `raw-dep.js`
- `raw-module1.js` (depends on `raw-dep.js`)

Library2 (depends on Library1) provides 1 raw module and owns the bundle config:
- `raw-module2.js` (depends on `raw-dep.js`)

Filters in raw section (with `resolve: true`):
- `raw-module2.js`
- `raw-module1.js`

Expected raw-module order:
- `raw-dep.js`
- `raw-module2.js`
- `raw-module1.js`

Maybe I'm missing something, but in this case I do not see a way to ensure that `raw-module2.js` comes before `raw-module1.js`, and sorting alphabetically would make it deterministic, but would always add module1 before module2.

codeworrior:
As I didn't immediately got the point of Matthias' (valid) scenario: 
There might be cases, where dependencies can't be declared, e.g.
* they would have to be declared in places, where they can't be declared (e.g. apps don't have a .library file)
* they would be "forward" declarations, creating a cycle (in Matthias' example, a dependency from a module of lib 1 to a module of lib 2 can'T be declared in lib 1 as it should not "know" lib 2.

There are other solutions for this problem (e.g. two "raw" sections"). But I agree that retaining the order of the filters as long as dependency order is honoured first, is a much simpler mechanism.

codeworrior (approved):

RandomByte (approved):

PR #1223, opened 2026-07-28
[FEATURE] libraryLessGenerator: Rewrite svg.less imports

stopcoder:
- The libraryLessGenerator processor already rewrites base.less imports from the UI5 resource path (sap/ui/core/themes/<theme>/base.less) to the theming-base-content path (Base/baseLib/<theme>/base.less) used by the Theme Designer. Extend this to also rewrite svg.less imports to Base/icons/<theme>/svg.less.

- The Theme Designer resolves imports against the theming-base-content structure. When a customer creates a custom theme, icon overrides are stored under Base/icons/<custom_theme>/. The parent theme's library.less must import from the same Base/icons/ path so that custom content layered on top takes effect via LESS "last definition wins" semantics.

- Extract a shared rewriteThemeDesignerImport() static method and drive both rewrites from a config array, so adding future rewrites is a one-liner.

JIRA: CPOUI5FRAMEWORK-812

### DIFF
--- a/test/lib/processors/libraryLessGenerator.js
+++ b/test/lib/processors/libraryLessGenerator.js
@@ -395,6 +395,64 @@ test.serial("LibraryLessGenerator: Rewrite for sap.ui.core base.less (from diffe
 	t.is(loggerStub.error.callCount, 0);
 });
 
+test.serial("LibraryLessGenerator: Rewrite for sap.ui.core svg.less", async (t) => {
+	const {LibraryLessGenerator, loggerStub} = t.context;
+
+	const input = `@import "../base/svg.less";\n` +
+		`@import "svg.less";`;
+	const expectedOutput = `${FILE_HEADER}\n\n` +
+	`@import "../../../../../../Base/icons/baseTheme/svg.less"; ` +
+		`/* ORIGINAL IMPORT PATH: "../base/svg.less" */\n\n` +
+	`@import "../../../../../../Base/icons/sap_horizon/svg.less"; ` +
+		`/* ORIGINAL IMPORT PATH: "svg.less" */\n`;
+
+	const fs = {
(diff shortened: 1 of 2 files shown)


test/lib/processors/libraryLessGenerator.js
```diff
@@ -395,6 +395,64 @@ test.serial("LibraryLessGenerator: Rewrite for sap.ui.core base.less (from diffe
...
+	const {LibraryLessGenerator, loggerStub} = t.context;
+
+	const input = `@import "../base/svg.less";\n` +
+		`@import "svg.less";`;
+	const expectedOutput = `${FILE_HEADER}\n\n` +
+	`@import "../../../../../../Base/icons/baseTheme/svg.less"; ` +
+		`/* ORIGINAL IMPORT PATH: "../base/svg.less" */\n\n` +
+	`@import "../../../../../../Base/icons/sap_horizon/svg.less"; ` +
+		`/* ORIGINAL IMPORT PATH: "svg.less" */\n`;
+
```
d3xter666, line 408: Is this going to break, if the default theme in UI5 changes?

As far as I undrestand this change, the resolution falls back to default theme's path and that's why `horizon` is included here. 

Is my assumption correct?
If it is, maybe rewriting the test to be more stable and expecting a pattern, than a string would be more stable in that case. (if we switch UI5's default theme in the future)

matz3: The processor runs for `sap_horizon` (see code below that calls `LibraryLessGenerator#generate`).
This is the reason why the theme name appears in the expected output.
There is not and (currently) should not be any hardcoded logic about default themes within any of our processors or tasks. Currently, the default theme is only a runtime concept.

d3xter666: Thanks!

matz3 (approved):

